PR #14, opened 2019-06-11
Make a copy of configuration being passed in

playground-julia:
Tests currently do not pass because the same config object is being used between the two bindings and the Content Type is being overridden. Changing the constructor to copy the config fixes this problem.

### DIFF
--- a/lib/bindings/http/binary_0_1.js
+++ b/lib/bindings/http/binary_0_1.js
@@ -1,7 +1,7 @@
 var axios = require("axios");
 
 function HTTPBinary(configuration){
-  this.config = configuration;
+  this.config = JSON.parse(JSON.stringify(configuration));
 
   this.config["headers"] = {
     "Content-Type":"application/cloudevents+json; charset=utf-8"
--- a/lib/bindings/http/binary_0_2.js
+++ b/lib/bindings/http/binary_0_2.js
@@ -2,7 +2,7 @@ var axios = require("axios");
 var empty = require("is-empty");
 
 function HTTPBinary(configuration){
-  this.config = configuration;
+  this.config = JSON.parse(JSON.stringify(configuration));
 
   this.config["headers"] = {
     "Content-Type":"application/cloudevents+json; charset=utf-8"
--- a/lib/bindings/http/structured_0_1.js
+++ b/lib/bindings/http/structured_0_1.js
@@ -1,7 +1,7 @@
 var axios = require("axios");
 
 function HTTPStructured(configuration){
-  this.config = configuration;
+  this.config = JSON.parse(JSON.stringify(configuration));
 
   this.config["headers"] = {
     "Content-Type":"application/cloudevents+json; charset=utf-8"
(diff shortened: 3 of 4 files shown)


lib/bindings/http/binary_0_1.js
```diff
@@ -1,7 +1,7 @@
 var axios = require("axios");
 
 function HTTPBinary(configuration){
-  this.config = configuration;
+  this.config = JSON.parse(JSON.stringify(configuration));
```
fabiojose: Please check out the [line #14](https://github.com/cloudevents/sdk-javascript/blob/77bad8d07f3c707446a5414d0a1e658c2752f8cf/lib/bindings/http/binary_0_1.js#L14). I suggest you remove this copy because it is useless.

lib/bindings/http/binary_0_2.js
```diff
@@ -2,7 +2,7 @@ var axios = require("axios");
 var empty = require("is-empty");
 
 function HTTPBinary(configuration){
-  this.config = configuration;
+  this.config = JSON.parse(JSON.stringify(configuration));
```
fabiojose: Check out the [line #15](https://github.com/cloudevents/sdk-javascript/blob/77bad8d07f3c707446a5414d0a1e658c2752f8cf/lib/bindings/http/binary_0_2.js#L15). The same way.

lib/bindings/http/structured_0_1.js
```diff
@@ -1,7 +1,7 @@
 var axios = require("axios");
 
 function HTTPStructured(configuration){
-  this.config = configuration;
+  this.config = JSON.parse(JSON.stringify(configuration));
```
fabiojose: Check out the [line #14](https://github.com/cloudevents/sdk-javascript/blob/77bad8d07f3c707446a5414d0a1e658c2752f8cf/lib/bindings/http/structured_0_1.js#L14). Same suggestion.

lib/bindings/http/structured_0_2.js
```diff
@@ -1,7 +1,7 @@
 var axios = require("axios");
 
 function HTTPStructured(configuration){
-  this.config = configuration;
+  this.config = JSON.parse(JSON.stringify(configuration));
```
fabiojose: Check out the [line #14](https://github.com/cloudevents/sdk-javascript/blob/77bad8d07f3c707446a5414d0a1e658c2752f8cf/lib/bindings/http/structured_0_2.js#L14).

lib/bindings/http/binary_0_1.js
playground-julia: I saw line #14. I don't think it's useless because the config object is being changed in line 6/7. That's what is causing npm test to fail...

fabiojose:
@playground-julia I've made some suggestions. Looking forward to your comments.

lib/bindings/http/binary_0_1.js
playground-julia: It might make more sense if you look at my other pull request first. https://github.com/cloudevents/sdk-javascript/pull/15

playground-julia:
Putting these changes in PR #15 and closing this. This was also not signed.

PR #16, opened 2019-06-11
Bug fix for example regarding emitting

playground-julia:
The README emitter example currently doesn't work because the word new is missing.


fabiojose:
@playground-julia Could you open an issue and link this PR there?

playground-julia:
Just did!

PR #21, opened 2019-06-22
Fixing initialization of binding

rinormaloku:
The following code:
```js
    const binding = Cloudevent.bindings["http-structured0.2"](config);
```
will return undefined. 

to create an instance we need to add the new keyword.


fabiojose:
Duplicated, see #16

Issue #33, opened 2019-08-22
Special handling of the "data" attribute not supported


joshwlewis:
In the v0.3 spec for JSON format section about [Special Handling of the "data" Attribute](https://github.com/cloudevents/spec/blob/v0.3/json-format.md#31-special-handling-of-the-data-attribute), it says: 

> First, if an implementation determines that the type of the data attribute is Binary or String, it MUST inspect the datacontenttype attribute to determine whether it is indicated that the data value contains JSON data.
> If the datacontenttype value is either "application/json" or any media type with a structured +json suffix, the implementation MUST translate the data attribute value into a JSON value, and set the data attribute of the envelope JSON object to this JSON value.

If I'm reading the spec correctly, it looks like the SDK should serialize stringified JSON in the data attribute. In other words, I would expect this to work:

```js
const v03 = require("cloudevents-sdk/v03");
const data = { "so": "scare" };
const ce = v03.event
  .id('4e4cfbd6-d542-4235-9e80-99d2df62abee')
  .source('com.example.whatever')
  .type('something')
  .dataContentType("application/json")
  .data(JSON.stringify(data)) // <-- stringified json data goes in
  .format();

const ceData = ce.getData();
console.log(ceData.so); // <-- this breaks, getData returned the stringified JSON, not an object.
```

I wrote a sample test (https://github.com/joshwlewis/sdk-javascript/commit/19cc9f3b0a081e84fd79dfae620d9435dc372a75), if it's more useful than my untested example. It fails with:

```
  CloudEvents Spec v0.3
    The Constraints check
      'data'
        1) should convert data with stringified json to a json object


  0 passing (8ms)
  1 failing

  1) CloudEvents Spec v0.3
       The Constraints check
         'data'
           should convert data with stringified json to a json object:
     AssertionError: expected '{"much":"wow"}' to equal { much: 'wow' }
      at Context.it (test/spec_0_3_tests.js:209:41)

```

fabiojose:
That references it's wrong, please, ignore it.

fabiojose:
@joshwlewis, after reading and talking with some fellows, we conclude that's right for incoming events, that needs to be unmarshaled. But, for the outcoming event, created by the `v03.event()...` the type will be as-is, no parsing at all.

If we parse a string to json in the `v03.event()..`, we will ending to marshall a json object, not a string.

That's makes sense to you?

fabiojose:
[...]

PR #35, opened 2019-09-01
To fix the 'data' special handling

fabiojose:
This PR has the intent to fix issue #33, by performing the special data handling described in the spec 0.3, [right here](https://github.com/cloudevents/spec/blob/v0.3/json-format.md#31-special-handling-of-the-data-attribute)


joshwlewis:
Awesome, thanks a bunch @fabiojose!

PR #43, opened 2019-11-18
Remove non-existing fromString from README API doc

danbev:
This commit removes the API documentation for the `fromString` Cloudevent
function as it does not seem to exist.


fabiojose:
Thank you @danbev

PR #49, opened 2020-04-02
deps: update old and remove unused dependencies

lance:
This commit updates all of the dependencies in package.json to
their most recent versions. It also removes dependencies that were
specified in package.json but not used - or only used in a
`require()` statement.

These changes have some ripple effects. Istanbul has not been
supported for some time, so it has been replaced with nyc.
The code coverage reporting tool from codacy has been updated
as well. This could not be tested without having the API token.

Finally, the CI job has been modified to run tests on Node.js
versions 10x and 12x. All older versions of Node.js are no longer
maintained.

Fixes: https://github.com/cloudevents/sdk-javascript/pull/50

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,12 @@
 language: node_js
 node_js:
-- 8
-- 7
-- 6
+- 12
+- 10
 cache: npm
 jobs:
   include:
   - stage: github release
-    node_js: '6'
+    node_js: '12'
     deploy:
       provider: releases
       api_key:
@@ -17,7 +16,7 @@ jobs:
         tags: true
 
   - stage: npm release
-    node_js: '6'
+    node_js: '12'
     deploy:
       provider: npm
       email: fabiojose@gmail.com
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 - Support for mTLS in v1.0 Binary and Structured Emitters: issue [#48](https://github.com/cloudevents/sdk-javascript/issues/48). Note that this fix is only valid for v1.0 and does not address the problem in v0.3 and below.
 
+### Removed
+
+- Removed support for Node.js 6, 7 and 8 - all of which are EOL at this point. Travis CI modified to test on Node.js 10 and 12.
+
 ## [1.0.0]
 
 ### Added
--- a/lib/bindings/http/emitter_binary.js
+++ b/lib/bindings/http/emitter_binary.js
@@ -1,5 +1,4 @@
-const axios = require("axios");
-const empty = require("is-empty");
+var axios = require("axios");
 
 const Constants = require("./constants.js");
 const defaults = {};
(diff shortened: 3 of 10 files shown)


lance:
Not sure why the DCO check didn't re-run. I amended my commit and pushed it with a signoff yesterday as you can see on my fork: https://github.com/lance/sdk-javascript/commit/201fc8b19527b2194a61c94f2bfe5c304fe4dba4.

lance:
@fabiojose just checking in - any thoughts on this?

fabiojose:
@lance take a look at my reviews. We are dropping the support for 6, 7 and 8, right?

lance:
@fabiojose yes - this commit bumps the tested Node.js versions to 10 and 12. All other versions of Node.js are [currently EOL](https://nodejs.org/en/about/releases/) (Node.js 8 may still be in maintenance mode, but it will EOL at the end of April). 

I don't see any reviews from you on this. 🤷

.travis.yml
```diff
@@ -1,13 +1,12 @@
 language: node_js
 node_js:
-- 8
```
fabiojose, line 3: People with node 8, 7 and 6 will be able to use cloudevent released with version 12?

lance: It's possible, but not guaranteed. I don't *think* the SDK uses language features that are not available in 6, 7 and 8. But I question the wisdom of even attempting to support a Node.js version that is end of life. Both 6 and 8 are EOL, and 7 was never an LTS version (no odd numbered versions of Node.js should be considered LTS ever).

fabiojose: Ok.

fabiojose:
@lance Can you add the entry in the CHANGELOG.md and fix that conflicting files?

.travis.yml
fabiojose: LGTM

lance:
@fabiojose requested changes have been made.

lance:
@fabiojose did you see my comment [here](https://github.com/cloudevents/sdk-javascript/pull/67#issuecomment-620869957)? The problem occurs during CI and seems to be an issue with the codacy script possibly not handling things well. I don't have credentials to publish codacy results or configure travis, so it's impossible for me to test this.

When I run locally I get this:

```console
**######################################################################## 100.0%
######################################################################## 100.0%
2020-04-29 12:27:25,617 ERROR            com.codacy Invalid configuration: Either a project token or an api token must be provided or available in an environment variable 
**
```

Note, that in this commit, the code coverage reporting changed [here](https://github.com/cloudevents/sdk-javascript/pull/49/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L8-R10). I noted this in the original PR text:

> The code coverage reporting tool from codacy has been updated
> as well. This could not be tested without having the API token.

I'll keep poking but I suspect I'll either need credentials for Travis and Codacy or someone who has those credentials will need to fix it.

The failure began with the Merge commit based on this PR. But as you can see, the PR itself passed the Travis build with no problems.

https://travis-ci.org/github/cloudevents/sdk-javascript/builds/677389800

So I'm happy to continue trying to figure out what went wrong, but not sure I can be successful without proper credentials.

fabiojose:
Seems all is ok.

PR #74, opened 2020-04-29
chore: add npm fix command

grant:
Since there is a `lint` command, there should be a way to fix easily.


grant:
The Travis failure is not due to changes in this PR.

lance:
I’m not sure this is really needed. The same result can be had by running `npm run lint — —fix`.

grant:
Where do we document how to run `npm run lint — —fix`? I didn't know about this.

lance:
@grant I'm not sure if it needs to be documented, since it's in the [npm documentation](https://docs.npmjs.com/cli/run-script).

> As of npm@2.0.0, you can use custom arguments when executing scripts. The special option -- is used by getopt to delimit the end of the options. npm will pass all the arguments after the -- directly to your script:

But perhaps adding it to the [contributor's guide](https://github.com/cloudevents/sdk-javascript/blob/master/CONTRIBUTING.md) would be a good place.

lance:
@grant you should rebase from `master` to pull in the Travis-CI code coverage fix. That will re-run the tests and you'll see a successful build.

grant:
Still would like to run a simple command like `fix` instead of `lint — —fix` which technically makes sense but is not intuitive.

Issue #76, opened 2020-04-29
Error when installing Express example


helio-frota:
```console
$ npm install
npm ERR! code 1
npm ERR! Command failed: git checkout v1
npm ERR! error: pathspec 'v1' did not match any file(s) known to git
npm ERR! 
```

I'm going to send a PR to fix and align that with the same approach used by TypeScript example.

helio-frota:
landed here: https://github.com/cloudevents/sdk-javascript/commit/bb8e0f9e0ca7aef00103d03f6071a648a9fab76d

PR #77, opened 2020-04-29
fix: Fix Express example installation

helio-frota:
Related to #76 

Now it uses the same approach as TypeScript example: https://github.com/cloudevents/sdk-javascript/blob/master/examples/typescript-ex/package.json#L28

### DIFF
--- a/examples/express-ex/package.json
+++ b/examples/express-ex/package.json
@@ -14,7 +14,7 @@
   "author": "fabiojose@gmail.com",
   "license": "Apache-2.0",
   "dependencies": {
-    "cloudevents-sdk": "github:cloudevents/sdk-javascript#v1",
+    "cloudevents-sdk": "~1.0.0",
     "express": "^4.17.1"
   }
 }


examples/express-ex/package.json
```diff
@@ -14,7 +14,7 @@
   "author": "fabiojose@gmail.com",
   "license": "Apache-2.0",
   "dependencies": {
-    "cloudevents-sdk": "github:cloudevents/sdk-javascript#v1",
+    "cloudevents-sdk": "1.0.0",
```
lance, line 17: Maybe it would be good to use a [patch or minor release filter](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept) here?

```suggestion
    "cloudevents-sdk": "~1.0.0",
```

helio-frota: Probably yes :+1:

helio-frota: Commit updated

PR #78, opened 2020-04-29
chore: fix CI code coverage publishing

lance:
The code coverage publishing in package.json works fine on non-ubuntu systems.
However, in Ubuntu `/bin/sh` is symlinked to `/bin/dash` which doesn't support
everything supported by `/bin/bash` - specifically `-o pipefail`.

This commit changes the code coverage script to explicitly use `bash` instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "test": "mocha test/**/*.js",
     "coverage": "nyc --reporter=lcov --reporter=text npm run test",
     "precoverage-publish": "npm run coverage",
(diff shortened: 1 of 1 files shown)


lance:
Note: coverage results are only published when a commit lands on `master`. For pull requests, the tests are run but coverage is not published. I wonder if this issue we are facing might indicate that not publishing coverage results on PRs is a problem. In fact, I can see how it would be desirable to publish code coverage reports for pull requests because you don't want coverage to go _down_ with a PR but you can't know that it will or not until after the merge.

lance:
FYI - in Codacy, it is possible to have the quality check fail if code coverage is reduced. Might be a good idea to do that.

PR #95, opened 2020-04-29
fix: Unhandled promise rejection in tests

helio-frota:


test/bindings/http/unmarshaller_0_3_tests.js
```diff
@@ -110,8 +110,12 @@ describe("HTTP Transport Binding Unmarshaller for CloudEvents v0.3", () => {
       // act and assert
       un.unmarshall(payload, headers)
         .then(() => { throw new Error("failed"); })
-        .catch((err) =>
-          expect(err.message).to.equal("invalid payload"));
+        .catch((err) => {
+          expect(err.message).to.equal("invalid payload");
+        })
+        .catch((err) => {
+          console.error(err);
```
lance, line 117: I think this should fail the test. If the message is not correct, it will throw. This seems to indicate that there's something wrong with the test itself. It's failing but not registering as a failure.

fabiojose:
@helio-frota have you take a look at @lance's requested changes?

helio-frota:
@fabiojose yeah I still need to understand the issue. I agree with @lance that probably the test has something wrong. 

But in other hand that test will be removed since it is related to spec 0.3
depending on https://github.com/cloudevents/sdk-javascript/issues/99

```js
it("Throw error when the event does not follow the spec 0.3", () => {
```

This PR is just to remove the Node warning:

```console
UnhandledPromiseRejectionWarning: AssertionError: expected 'failed' to equal 'invalid payload'
```

When we deal with another catch block, we can see clearly the result:

```js
{
  message: "expected 'failed' to equal 'invalid payload'",
  showDiff: true,
  actual: 'failed',
  expected: 'invalid payload'
}
```

I think it is better to close this PR and wait for that warning message to be removed without a discarded  effort. :smiley: 

Please, let me know what do you think :+1:

lance:
@helio-frota I did some experimenting with this over the weekend. These tests are particularly problematic in a way that the 1.0 version of the tests is not - specifically because the 0.3 unmarshalling is async. Any of the unmarshalling tests for 0.3 will exhibit this same behavior. When the call to `expect()` fails, Chai throws an exception. But since this is happening asynchronously, it's not caught by the test framework and won't be registered as a failure. I think to effectively test this, something like [Chai As Promised](https://www.chaijs.com/plugins/chai-as-promised/) would need to be used.

I think we need to determine [whether 0.3 is going to be removed](https://github.com/cloudevents/sdk-javascript/pull/110). If not, we need to fix this for all of the 0.3 unmarshaling tests. If 0.3 goes away, then this problem magically disappears on `master` but I still think it should be fixed on the `v1.x-backport` branch.

helio-frota:
@lance yeah, I agree with that, I'm favor to wait for #110 decision.
I'm going to keep the PR open for about 72 hours, then revisit and see what can be done (if nobody does it before -- which will be good if that happen of course)

Sound good ?

helio-frota:
Seems good now and able to be merged :+1:

test/bindings/http/unmarshaller_0_3_tests.js
helio-frota: PR updated, please take a look when you have a time. :+1:

test/bindings/http/unmarshaller_0_3_tests.js
```diff
@@ -108,10 +108,11 @@ describe("HTTP Transport Binding Unmarshaller for CloudEvents v0.3", () => {
       const un = new Unmarshaller();
 
       // act and assert
-      un.unmarshall(payload, headers)
-        .then(() => { throw new Error("failed"); })
-        .catch((err) =>
-          expect(err.message).to.equal("invalid payload"));
+      try {
+        await un.unmarshall(payload, headers);
```
lance, line 112: @helio-frota you should fail the test after this line. The expectation is that the unmarshalling should throw if the incoming message does not conform to spec. As the test is now, that isn't happening. We never enter the catch block below. You can change that `.to.equal()` call to be `.to.equal('tacos')` and the test still passes.

helio-frota: OOps you right! I was only focused on solve the unhandled promise exception.
Thanks for the review, commit updated.

helio-frota: It still not good btw..
I'm trying to understand what is happening...

helio-frota:
Ok I found that `spec.check(payload)` was never called.
I think it is good for this PR which initially was addressing the unhandled promise warning.

If the `spec.check()` is not being called on the right place and/or other thing. Then in this case it will be good to create a new issue to fix that. (or to discard that in case v0.3 removal)

lance:
@helio-frota as we keep nibbling around the edges of what is a bigger problem, I'm starting to think that the right move here is to modify the 0.3 unmarshaling API so that it's no longer asynchronous and matches the API exposed for the 1.0 protocol. There is no reason that this API should be async.

Maybe the best plan of attack is to open a new PR that addresses that. WDYT?

helio-frota:
@lance I agree with that. (Not saying that I'm going to send a PR soon afterwards :smile: )
Also not sure if I'm currently able to do that size of refactor on this repo.

But anyway related to this specific issue I think we are good, the `spec.check()` was not executed and that was causing the issue related to the `expect`. And the unhandled promise warning is another issue.

lance:
Closing - superseded by https://github.com/cloudevents/sdk-javascript/pull/126

PR #101, opened 2020-04-30
refactor: use CloudEvents not cloudevents everywhere

### LINKED ISSUE #100: Spell CloudEvent Correctly and Consistently
There are a few places where the term `CloudEvent` is spelled and named in different ways:
- `event`
- `cloudevent`
- `Cloudevent`
- `CloudEvent`

We should spell the term the same and not have multiple options for referring to the same thing.

grant:
Uses the same capitalization for CloudEvents everywhere.
Fixes #100.

### DIFF
--- a/v03/index.js
+++ b/v03/index.js
@@ -1,4 +1,4 @@
-const Cloudevent = require("../lib/cloudevent.js");
+const CloudEvent = require("../lib/cloudevent.js");
 const Spec = require("../lib/specs/spec_0_3.js");
 const StructuredHTTPEmitter =
   require("../lib/bindings/http/emitter_structured.js");
@@ -12,8 +12,8 @@ const BinaryHTTPReceiver =
 
 const HTTPUnmarshaller = require("../lib/bindings/http/unmarshaller_0_3.js");
 
-function event() {
-  return new Cloudevent(Spec);
+function newEvent() {
+  return new CloudEvent(Spec);
 }
 
 module.exports = {
@@ -23,7 +23,6 @@ module.exports = {
   BinaryHTTPEmitter,
   BinaryHTTPReceiver,
   HTTPUnmarshaller,
-  Cloudevent,
-  CloudEvent: Cloudevent,
-  event
+  CloudEvent: newEvent,
+  event: newEvent
 };
--- a/v1/index.js
+++ b/v1/index.js
@@ -1,4 +1,4 @@
-const Cloudevent = require("../lib/cloudevent.js");
+const CloudEvent = require("../lib/cloudevent.js");
 const Spec = require("../lib/specs/spec_1.js");
 
 const StructuredHTTPEmitter =
@@ -12,8 +12,8 @@ const StructuredHTTPReceiver =
 const BinaryHTTPReceiver =
   require("../lib/bindings/http/receiver_binary_1.js");
 
-function event() {
-  return new Cloudevent(Spec);
+function newEvent() {
+  return new CloudEvent(Spec);
 }
 
 module.exports = {
@@ -22,7 +22,6 @@ module.exports = {
   BinaryHTTPEmitter,
   StructuredHTTPReceiver,
   BinaryHTTPReceiver,
-  Cloudevent: event,
-  CloudEvent: event,
-  event
+  CloudEvent: newEvent,
+  event: newEvent
 };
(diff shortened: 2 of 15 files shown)


helio-frota:
:+1: 
It would be good to apply the same camelCase for variable names as well, in another PR.

```js
from
const cloudevent
to
const cloudEvent
```

v03/index.js
```diff
@@ -23,7 +23,6 @@ module.exports = {
   BinaryHTTPEmitter,
   BinaryHTTPReceiver,
   HTTPUnmarshaller,
-  Cloudevent,
```
lance, line 26: This is a breaking change to the API. I think we need to branch a `v1.x` line before doing this, or simply leave this line in for backwards compat.

v03/index.js
```diff
@@ -12,8 +12,8 @@ const BinaryHTTPReceiver =
 
 const HTTPUnmarshaller = require("../lib/bindings/http/unmarshaller_0_3.js");
 
-function event() {
-  return new Cloudevent(Spec);
+function newEvent() {
+  return new CloudEvent(Spec);
 }
```
lance, line 17: If we are exporting the constructor, why even have this function?

v1/index.js
```diff
@@ -12,8 +12,8 @@ const StructuredHTTPReceiver =
 const BinaryHTTPReceiver =
   require("../lib/bindings/http/receiver_binary_1.js");
 
-function event() {
-  return new Cloudevent(Spec);
+function newEvent() {
+  return new CloudEvent(Spec);
 }
```
lance, line 17: Again - if we are exporting the ctor...

v1/index.js
```diff
@@ -22,7 +22,6 @@ module.exports = {
   BinaryHTTPEmitter,
   StructuredHTTPReceiver,
   BinaryHTTPReceiver,
-  Cloudevent: event,
```
lance, line 25: Breaking change...

v1/index.js
fabiojose: @lance If I am correct, spec 0.2 had SDK conformance to expose a method event() to create new event instances. For now, it's just to not break the sdk API

v03/index.js
grant: Yes, this is a breaking change. I can change the commit message accordingly.
This would be a `2.0`, we don't need separate git branches for breaking changes, just major version changes.

v03/index.js
grant: I didn't attempt to change functionality, just the name in this PR.
I don't want functionality change beyond the name changes.

v1/index.js
grant: I agree. However, this code was here before. I'm just changing the name.

v1/index.js
grant: Yes. This is a PR with a breaking change. We'll cut a new version.

v03/index.js
lance: @grant I'm not sure you understand my proposal. I am not suggesting that we have branches for breaking changes.

v03/index.js
lance: Seems like an arbitrary distinction. I'd argue that making breaking changes as noted above is a change in functionality.

v03/index.js
grant: OK, let's talk about the breaking change issue separately and reconvene here.

grant:
I'm still in favor of merging this PR. If we think a part of the PR should be changed, please request it.

lance:
Yes - let's land it at some point. Let's just make sure that we know what we are doing with regard to breaking changes before we do that. OK?

grant:
> Yes - let's land it at some point. Let's just make sure that we know what we are doing with regard to breaking changes before we do that. OK?

I would like to not say "at some point". I don't understand what the blocker is here as the `v1.x.y` system sounds fine.

lance:
@grant Literally 20 minutes ago you said in Slack that you wanted something different with branch management. Is it fine now?

grant:
> @grant Literally 20 minutes ago you said in Slack that you wanted something different with branch management. Is it fine now?

I'm still understanding the backporting feature. It seems fine in terms of branch management so long as we can make progress with future PRs like this that make breaking changes.

grant:
Thanks for your patience and responsiveness @lance.

Issue #97, opened 2020-04-30
Enforce ES6 with Lint Rules


grant:
Modern JavaScript doesn't use `var`. Instead use `let` or `const`.

We can enforce modern JavaScript through lint rules such as ESLint.

PR #104, opened 2020-04-30
chore: update eslint rules to disallow var usage

lance:
Enforce the use of `let` and `const` by using elsint rules.
When creating the eslint configuration, I had assumed that
`extends: eslint:recommended` would have covered this, but
apparently not!

Existing usage of `var` fixed with `npm run lint -- --fix`.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/97

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,6 +6,7 @@
     "mocha": true
   },
   "rules": {
+    "no-var": "error",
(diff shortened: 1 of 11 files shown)


fabiojose:
LGTM

Issue #48, opened 2020-03-24
getting error using httpsAgent in http binding emitter


nan-xiong:
we need to use mTLS authentication to a https endpoint with certificate and key when submitting cloud events.  
We got the following error:
"The "options.agent" property must be one of type Agent-like Object, undefined, or false." 
when using the following code:
`var options = {
        method: "POST",
        url: requestUrl,
        httpsAgent: new https.Agent({
          cert: fs.readFileSync(cf_cert),
          key: fs.readFileSync(cf_key)
        })
      };`
`var binding = new cloudeventsSDK.StructuredHTTPEmitter(options);
    binding.emit(ce)
      .then(response => {
        callback(null);
      }).catch(err => {
        callback(err);
      });`

emitter_structured.js (possibly in other emitters as well) line #6 & line#23 causing the issue, it assumes the configuration is a JSON object, but in the case of httpsAgent, the underlying module requires httpsAgent attribute of type 'Agent-like' object.

![image](https://user-images.githubusercontent.com/62610243/77455025-920ca880-6db6-11ea-95fa-c570d7060538.png)

PR #53, opened 2020-04-08
fix: support mTLS in 1.0 Binary and Structured emitters

lance:
This commit modifies both of the 1.0 emitters so that they may
accept typed objects as a part of the configuration. When using
mTLS in Node, you need to provide an `Agent` to the underlying
HTTP handler. In this case, Axios will pass this object along to
Node.js when it is provided.

Note: This does not address the issue in 0.3 or below.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/48

Signed-off-by: Lance Ball <lball@redhat.com>

PR #106, opened 2020-04-30
fix: [backport] support mTLS in 1.0 Binary and Structured emitters

lance:
This commit modifies both of the 1.0 emitters so that they may
accept typed objects as a part of the configuration. When using
mTLS in Node, you need to provide an `Agent` to the underlying
HTTP handler. In this case, Axios will pass this object along to
Node.js when it is provided.

Landed in: https://github.com/cloudevents/sdk-javascript/commit/3a063d72451d1156df8fe9c3499ef1e81e905060
Pull Request: https://github.com/cloudevents/sdk-javascript/pull/53

Fixes: https://github.com/cloudevents/sdk-javascript/issues/48

### DIFF
--- a/lib/bindings/http/emitter_binary.js
+++ b/lib/bindings/http/emitter_binary.js
@@ -2,58 +2,48 @@ var axios = require("axios");
 var empty = require("is-empty");
 
 const Constants = require("./constants.js");
+const defaults = {};
+defaults[Constants.HEADERS] = {};
+defaults[Constants.HEADERS][Constants.HEADER_CONTENT_TYPE] = Constants.DEFAULT_CONTENT_TYPE;
 
 function BinaryHTTPEmitter(config, headerByGetter, extensionPrefix){
-  this.config = JSON.parse(JSON.stringify(config));
+  this.config = Object.assign({}, defaults, config);
   this.headerByGetter = headerByGetter;
   this.extensionPrefix = extensionPrefix;
-
-  this.config[Constants.HEADERS] =
-    (!this.config[Constants.HEADERS]
-      ? {}
-      : this.config[Constants.HEADERS]);
-
-  // default is json
-  if(!this.config[Constants.HEADERS][Constants.HEADER_CONTENT_TYPE]){
-    this.config[Constants.HEADERS][Constants.HEADER_CONTENT_TYPE] =
-      Constants.DEFAULT_CONTENT_TYPE;
-  }
 }
 
-BinaryHTTPEmitter.prototype.emit = function(cloudevent) {
-  // Create new request object
-  var _config = JSON.parse(JSON.stringify(this.config));
-
-  // Always set stuff in _config
-  var _headers = _config[Constants.HEADERS];
+BinaryHTTPEmitter.prototype.emit = function (cloudevent) {
+  const config = Object.assign({}, this.config);
+  const headers = Object.assign({}, this.config[Constants.HEADERS]);
 
   Object.keys(this.headerByGetter)
     .filter((getter) => cloudevent[getter]())
     .forEach((getter) => {
-      let header = this.headerByGetter[getter];
-      _headers[header.name] =
+      const header = this.headerByGetter[getter];
+      headers[header.name] =
         header.parser(
           cloudevent[getter]()
         );
     });
 
   // Set the cloudevent payload
-  let formatted = cloudevent.format();
+  const formatted = cloudevent.format();
   let data = formatted.data;
   data = (formatted.data_base64 ? formatted.data_base64: data);
 
-  _config[Constants.DATA_ATTRIBUTE] = data;
-
   // Have extensions?
-  var exts = cloudevent.getExtensions();
+  const exts = cloudevent.getExtensions();
   Object.keys(exts)
     .filter((ext) => Object.hasOwnProperty.call(exts, ext))
     .forEach((ext) => {
-      _headers[this.extensionPrefix + ext] = exts[ext];
+      headers[this.extensionPrefix + ext] = exts[ext];
     });
 
-  // Return the Promise
-  return axios.request(_config);
-};
+  config[Constants.DATA_ATTRIBUTE] = data;
+  config.headers = headers;
+
+    // Return the Promise
+  return axios.request(config);
+ };
 
 module.exports = BinaryHTTPEmitter;
--- a/lib/bindings/http/emitter_structured.js
+++ b/lib/bindings/http/emitter_structured.js
@@ -1,30 +1,23 @@
 var axios = require("axios");
 
 const Constants = require("./constants.js");
+const defaults = {};
+defaults[Constants.HEADERS] = {};
+defaults[Constants.HEADERS][Constants.HEADER_CONTENT_TYPE] = Constants.DEFAULT_CE_CONTENT_TYPE;
 
 function StructuredHTTPEmitter(configuration){
-  this.config = JSON.parse(JSON.stringify(configuration));
-
-  this.config[Constants.HEADERS] =
-    (!this.config[Constants.HEADERS]
-      ? {}
-      : this.config[Constants.HEADERS]);
-
-  if(!this.config[Constants.HEADERS][Constants.HEADER_CONTENT_TYPE]){
-    this.config[Constants.HEADERS][Constants.HEADER_CONTENT_TYPE] =
-      Constants.DEFAULT_CE_CONTENT_TYPE;
-  }
+  this.config = Object.assign({}, defaults, configuration);
 }
 
-StructuredHTTPEmitter.prototype.emit = function(cloudevent) {
-  // Create new request object
-  var _config = JSON.parse(JSON.stringify(this.config));
-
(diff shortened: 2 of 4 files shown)


lib/bindings/http/emitter_structured.js
```diff
@@ -1,30 +1,23 @@
...
   // Set the cloudevent payload
-  _config[Constants.DATA_ATTRIBUTE] = cloudevent.format();
+  this.config[Constants.DATA_ATTRIBUTE] = cloudevent.format();
 
   // Return the Promise
-  return axios.request(_config);
+  return axios.request(this.config).then(response => {
+    delete this.config[Constants.DATA_ATTRIBUTE];
+    return response;
+  });
```
grant, line 20: This is very confusing.

Why don't we just do this?

```js
const res = await axios.request(this.config);
delete this.config[Constants.DATA_ATTRIBUTE];
return res;
```

The outer function needs to be async of course.

lance:
The Codacy failure is due to [this failure](https://app.codacy.com/manual/cloudevents/sdk-javascript/pullRequest?prid=5451944):

> Seems like the v1.x.y is not being analyzed. To review this pull request enable the branch in the repository settings.

lib/bindings/http/emitter_structured.js
```diff
@@ -1,30 +1,23 @@
 var axios = require("axios");
 
 const Constants = require("./constants.js");
+const defaults = {};
```
grant, line 4: Why are we creating an empty object and then adding properties to it?

```js
const defaults = {
  [Constants.HEADERS]: {
    [Constants.HEADER_CONTENT_TYPE]: Constants.DEFAULT_CE_CONTENT_TYPE
  }
};
```

lib/bindings/http/emitter_structured.js
```diff
@@ -1,30 +1,23 @@
...
-  this.config[Constants.HEADERS] =
-    (!this.config[Constants.HEADERS]
-      ? {}
-      : this.config[Constants.HEADERS]);
-
-  if(!this.config[Constants.HEADERS][Constants.HEADER_CONTENT_TYPE]){
-    this.config[Constants.HEADERS][Constants.HEADER_CONTENT_TYPE] =
-      Constants.DEFAULT_CE_CONTENT_TYPE;
-  }
+  this.config = Object.assign({}, defaults, configuration);
```
grant, line 9: Why are we using `Object.assign` when we can just use the native JavaScript object spread operator?

```js
this.config = {...defaults, ...configuration};
```

v1/index.js
```diff
@@ -22,5 +22,7 @@ module.exports = {
   BinaryHTTPEmitter,
   StructuredHTTPReceiver,
   BinaryHTTPReceiver,
+  Cloudevent: event,
```
grant, line 25: Can we not add `Cloudevent` and just use `CloudEvent`? There is no reason for an alias.

test/http_binding_1.js
```diff
@@ -117,6 +132,21 @@ describe("HTTP Transport Binding - Version 1.0", () => {
...
+    it('works with mTLS authentication', () => {
+      const event = new BinaryHTTPEmitter({
+        method: 'POST',
+        url: `${webhook}/json`,
+        httpsAgent: new https.Agent({
+          cert: 'some value',
+          key: 'other value'
+        })
+      })
+      return event.emit(cloudevent).then(response => {
```
grant, line 144: This can also be converted to ES6.

lib/bindings/http/emitter_binary.js
```diff
@@ -2,58 +2,48 @@ var axios = require("axios");
...
+      headers[this.extensionPrefix + ext] = exts[ext];
     });
 
-  // Return the Promise
-  return axios.request(_config);
-};
+  config[Constants.DATA_ATTRIBUTE] = data;
+  config.headers = headers;
+
+    // Return the Promise
```
grant, line 45: Formatting.

lance:
@grant this is a backport of an existing commit, cherry picked onto the v1.x.y branch. I think it should remain as the original commit. This is an example.

lib/bindings/http/emitter_structured.js
lance: This change introduces a race condition.

lib/bindings/http/emitter_structured.js
lance: You introduced new syntax to me. I wasn't aware that you could bracket a nested property as a field name.

lib/bindings/http/emitter_structured.js
lance: The commit happened before we dropped support for early versions of Node.js so this would not have passed CI. It's a reasonable change for master now that we've dropped 6 and 8.

test/http_binding_1.js
lance: Again - this was prior to dropping Node 6 and 8 support.

lib/bindings/http/emitter_structured.js
grant: No, where's the race? That's not how Node executes.

lib/bindings/http/emitter_structured.js
grant: OK. This is a feature of JavaScript ES2015. Here's a reference:
https://stackoverflow.com/a/19837961

lib/bindings/http/emitter_structured.js
grant: I'm not sure of the CI system yet but Node versions < 10 were end of life Dec 31, 2019. Completely understandable, but we should only use non-deprecated versions of Node in our CI.

grant:
> @grant this is a backport of an existing commit, cherry picked onto the v1.x.y branch. I think it should remain as the original commit. The whole entire idea is that this 1.x branch get the exact commits that landed on master. If you want to change syntax to ES6 on master that's great. We can chose to backport that or not.

OK. So we're basically just freezing a version of master and setting it to this branch?

In that case, that's fine, and we should indicate the PR is meant for that. We probably don't need a review if this is just mirroring `master`.

lance:
> OK. So we're basically just freezing a version of master and setting it to this branch?

Sort of. Here I used `git cherry-pick` to apply a single commit from master to the `v1.x.y` branch. This results in the bug fix applied in the commit on `master` being backported to the `v1.x.y` branch. The idea is that we don't want _all_ of the commits from master landing in `v1.1.0` or any other subsequent release of the 1.x line. We just want the commits that are important. No new features, no breaking changes. Just bug fixes, etc.

I understand that there aren't a lot of people using this module right now. But, there are _some_ people using it, I'm sure. And those people are all on 1.0.0 right now. If our 2.0.0 release is going to have major breaking changes (say the API is completely changed - not using builder patter for example), we can't expect people to immediately rewrite their code to accommodate our new release. HOWEVER, they still should have the bug fixes and updates that keep their application working correctly through patch/minor version level releases of the 1.x line. That's what this whole branch/backport thing is about.

lance:
> In that case, that's fine, and we should indicate the PR is meant for that. We probably don't need a review if this is just mirroring master.

The `v1.x.y-backport` label is the indicator. As well as the two lines in the PR description,

> Landed in: 3a063d7
> Pull Request: #53

"Landed in: xxxxx" indicates what commit this cherry pick originally landed in on `master`.
"Pull Request: #53" indicates what pull request had the original fix.

Backporting isn't always easy and doesn't always apply cleanly. It's not as simple as "mirroring master" - it's taking a single commit from the master branch and applying it to the backport branch. As you are aware, the velocity of change in the repo has increased quite a bit. It's easy to get conflicts when backporting and you need to take care with what you are doing. I think it's important to still have approvals for these kinds of PRs. They shouldn't just be allowed to land without review.

lib/bindings/http/emitter_structured.js
fabiojose: I vote for `then` notation.

lib/bindings/http/emitter_structured.js
fabiojose: I did not know either! Looks good

v1/index.js
fabiojose, line 25: Good!

lib/bindings/http/emitter_structured.js
lance: @grant try this in your browser https://gist.github.com/lance/ff1bf6d2da191ad164257349486c4c95

lib/bindings/http/emitter_structured.js
lance: @grant yes, I am aware. That's why I updated CI last month to only use 10 & 12. https://github.com/cloudevents/sdk-javascript/commit/aa2cef6e3cb3a399afbea24006cc83d8d70af920#diff-354f30a63fb0907d4ad57269548329e3

helio-frota: @lance I've created this #119 but I don't think a new PR is necessary maybe if you and the others agree with that practice you can update your PR instead wdyt ?

lance: @helio-frota I think submitting a separate PR for that is fine.

PR #108, opened 2020-05-01
chore: es6 unmarshaller

grant:


lib/bindings/http/unmarshaller.js
```diff
@@ -47,32 +39,39 @@ function resolveBindingName(payload, headers) {
...
-      validateArgs(payload, headers);
+  unmarshall(payload, headers) {
+    return new Promise((resolve, reject) => {
+      try {
+        if (!payload) {
+          throw new TypeError("payload is null or undefined");
+        }
+        if (!headers) {
+          throw new TypeError("headers is null or undefined");
+        }
```
lance, line 55: I like that you have removed the `validateArgs()` function, but I don't think there's actually a reason to throw here. A simpler version of what you have might look like this. 

```suggestion
  unmarshall(payload, headers) {
    return new Promise((resolve, reject) => {
      if (!payload) {
        return reject(new TypeError("payload is null or undefined"));
      }
      if (!headers) {
        return reject(new TypeError("payload is null or undefined"));
      }
```
I don't think the try/catch should be there at all, actually. Its only purpose is to catch this exception and then reject the promise. If something else, e.g. `resolveBindingName()` were to throw, the caller's catch block on the thennable will handle it.

grant: Thanks. I agree, but I'm just trying to refactor without modifying the functionality.
I propose we just approve this PR and can see about removing the existing try/catches here and elsewhere.

lance: OK - just to be clear though, this suggested modification isn't changing functionality. It's just a minor optimization similar to your removal of the `validateArgs()` function.

grant: OK, I can commit your suggested change but I think it's going to create a bad PR, where it was an OK PR before.

grant:
I added the suggested changes after you insisted, but they break the build.

lance:
You could have made the change locally and easily seen the syntax error and then pushed them. Were you trying to prove a point? I didn't insist. I just suggested.

grant:
> You could have made the change locally and easily seen the syntax error and then pushed them. Were you trying to prove a point? I didn't insist. I just suggested.

The suggested change doesn't just introduce format errors which are easy to fix. It introduces a rogue "catch" statement. I don't understand what we're trying to do. I'm just trying to change the unmarshaller to es6, not change the file in this small PR which might require the tests to change.

Can we please approve the PR assuming we will revert the last commit?

grant:
There's also an error with a repeated "payload is null or undefined" error.

grant:
OK, I fixed the error. The tests now pass locally.

lance:
Grant, in a spirit of collaboration, I merely made a suggestion. Your veiled hostility is unwelcome.

grant:
OK, I just don't want suggestions of breaking the PR as it was perfectly fine before. There were 2 issues with the suggestion. It's hard enough to make PRs to this repo as is.

lance:
And yet you continue with the finger pointing.

grant:
I'm trying to fix the DCO error, I think the commit from GitHub with the suggested change doesn't isn't signed off and is breaking DCO.

Issue #99, opened 2020-04-30
Recommend dropping support for 0.3 in Favor of 1.0 (and 2.0)


grant:
I recommend we drop support for 0.3 events and only support the major releases of the CloudEvent spec. 0.3 is not a major release. There is no other major release. [See releases](https://github.com/cloudevents/spec/releases).

See go for example (v1 and v2):
https://github.com/cloudevents/sdk-go

> Each SDK supports the latest(N), and N-1, major releases of the CloudEvent spec*.
> Within the scope of a major release, only support for the latest minor version is needed.

If there's still debate as to the exact definition of this, but I propose we talk to the other SDK folks if there's really.

I don't think there are many users of this library that it should be a problem.

---

It would be dramatically easier to refactor this library and improve the experience if 0.3 wasn't here.

helio-frota:
That makes sense. :+1: 

"Better code is no-code"... without v0.3 I don't need to bother with a failing test anymore: https://github.com/cloudevents/sdk-javascript/pull/95

helio-frota:
npm has no support of download counts per version.

So we can only speculate on that:

Comparing downloads from ~6 months https://www.npmjs.com/package/cloudevents-sdk/v/0.3.2
https://npm-stat.com/charts.html?package=cloudevents-sdk&from=2019-10-01&to=2020-04-29

And looking the public projects that are using cloudevents-sdk:

https://github.com/micheleangioni/node-messagebrokers/blob/master/package.json#L52
https://github.com/boson-project/faas-js-runtime/blob/master/package.json#L22
https://gitlab.com/hahnpro/flow-sdk/-/blob/master/package.json#L32
https://github.com/joshwlewis/neverred-function/blob/master/package.json#L18

Some of repositories listed on npm as dependents are deprecated and/or 404.

/cc @lance

lance:
Just to be clear. There is no 2.0 version of the spec. The only two versions that are expected to be supported are 1.0 and (semi-optionally) 0.3. FWIW, the Go module does support both of these versions.

> [Note: Supported CloudEvents specification: [0.3, 1.0].](https://github.com/cloudevents/sdk-go/blob/master/README.md#working-with-cloudevents)

So even though I said yesterday that I agreed dropping 0.3 would make refactoring easier. I'm still on the fence about it.

grant:
> So even though I said yesterday that I agreed dropping 0.3 would make refactoring easier. I'm still on the fence about it.

Yeah, I can see that.

[...]

PR #110, opened 2020-05-01
Drop Support for CloudEvents 0.3

grant:
This PR drops support for 0.3 CloudEvents.
The main discussion of whether this should be done should continue in the issue (https://github.com/cloudevents/sdk-javascript/issues/99).


lance:
So, this has been discussed ad nauseum in https://github.com/cloudevents/sdk-javascript/issues/99 and on Slack. I am opposed to this change until the 2.0 version of the CE specification has been released. I currently have code that receives Cloud Events from various, unknown sources - some of which are still sending using the 0.3 protocol. I do not want to have this code fail if these event producers are still producing 0.3 events. Until the 2.0 version is released (or release candidate) and there is an implementation of that protocol I think 0.3 should remain.

The [SDK requirements doc](https://github.com/cloudevents/spec/blob/master/SDK.md#contribution-acceptance) says this about version compatibility.

> * Each SDK has active points of contact.
> * Each SDK supports the latest(N), and N-1, major releases of the CloudEvent spec*.
> * Within the scope of a major release, only support for the latest minor version is needed.
> * Support for release candidates is not required, but strongly encouraged.
> * Note: v1.0 is a special case and it is recommended that as long as v1.0 is the latest version, SDKs should also support v0.3.

I don't really see the justification for completely dropping support and breaking things for existing users, contrary to recommendations. We will need to continue supporting multiple protocol versions down the line, so this to me seems short sighted.

helio-frota:
> SDKs **_should_** also support v0.3.

That seems to be the situation that needs justification according to the [RFC](https://tools.ietf.org/html/rfc2119#section-3) which is followed by the [spec](https://github.com/cloudevents/spec/blob/master/spec.md#notational-conventions)

> So, this has been discussed ad nauseum in #99 and on Slack. I am opposed to this change until the 2.0 version of the CE specification has been released. I currently have code that receives Cloud Events from various, unknown sources - some of which are still sending using the 0.3 protocol. I do not want to have this code fail if these event producers are still producing 0.3 events. Until the 2.0 version is released (or release candidate) and there is an implementation of that protocol I think 0.3 should remain.

I think that is a valid justification so :-1:  for dropping 0.3 now.

Issue #85, opened 2020-04-29
Discussion: Version Branches


lance:
Currently the repository is tagged for each prerelease, mirroring the spec versions as they progressed. However, there are no release branches. This was probably fine prior to 1.0. But now that 1.0 is out, a `v1.x` branch should be created and maintained so that fixes can be backported to previous versions as we continue to move forward with changes in the repository.

I suggest the following:

* Create a `v1.x` branch off of the `v1.0.0` tag.
* Create a `v1.x-backport` label which can be used to label any PRs that have landed since the 1.0.0 release, indicating they should be considered for backporting to the `v1.x` branch.
* As appropriate, create patch releases off of the `v1.x` branch, tagging the repo  on the `v1.x` branch with `v1.0.1`, for example, for the next patch release.
* As appropriate, create minor releases off of the `v1.x` branch, tagging the repo on the `v1.x` branch with `v1.1.0`, for example, for the next minor release.
* If needed create a `v1.1.x` branch for minor release patch updates.
* As appropriate, create patch releases off of the `v1.1.x` branch, tagging that branch with `v1.1.1`, for example, for the next patch release in this major/minor combination. 
* Repeat for 2.0.0 and up on their own branches.

This is related to the question raised in https://github.com/cloudevents/sdk-javascript/issues/72 where the contributor's guide talks about using the `develop` branch. That hasn't been done since the 1.0.0 release. We need to decide if we want to move forward with this process. If so, then the `develop` branch needs to be updated with all recent commits, and set as the primary branch in the repository settings. This setting will ensure that by default, pull requests from contributors are created to land on the `develop` branch.

My personal feeling is that the `develop` branch is mostly redundant with `master` and we don't really need to do that. If contributors can manage rebasing their pull requests with updates from `master` regularly, this shouldn't be a problem. Though I think this does add some pressure to land PRs quickly. What are other's thoughts on this?

fabiojose:
@lance I am with you lets remove the `develop` and works with `master`. As you said, no activity landed to `develop` for a long time, then it useless to maintain this branch.

lance:
@fabiojose great - I'll leave time for others to comment and will start a PR with some docs around this (or whatever is eventually agreed on in this discussion).

grant:
This is a really long issue, so I'm sorry if I'm misunderstanding something.

I'm in favor of removing `develop` and working with `master`.

[...]

PR #122, opened 2020-05-05
chore: Modify CI to also build backport branch(es)

helio-frota:
Signed-off-by: Helio Frota <00hf11@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language: node_js
 node_js:
-- 12
+- lts/*
 - 10
 cache: npm
 jobs:
   include:
   - stage: github release
-    node_js: '12'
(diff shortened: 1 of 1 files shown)


lance:
Adding the label `stalled` here, as this may not be needed depending on the outcome of https://github.com/cloudevents/sdk-javascript/issues/85. We are currently planning a 2.0.0 release without plans for backporting anything to 1.0.0. Assuming a steady release cadence can be maintained, we may choose to support no prior versions once a new version is released. Stating here, for the record, I think that's a bad idea.

lance:
@helio-frota do you mind changing this to just handle the `lts` issue and removing the build of backport branches? (but keep master and pull request builds active). Thanks!

PR #126, opened 2020-05-05
feat(unmarshaller)!: remove asynchronous 0.3 unmarshaller API

lance:
This commit removes the unnecessary use of Promises in the 0.3 unmarshaller.
There was actually no asynchronous activity happening in that function, so
there was no need to deal with Promises, and as a result testing was made
much more difficult.

This is a breaking change.

Fixes: https://github.com/cloudevents/sdk-javascript/pull/95

### DIFF
--- a/test/bindings/http/unmarshaller_0_3_tests.js
+++ b/test/bindings/http/unmarshaller_0_3_tests.js
@@ -10,8 +10,7 @@ const schemaurl = "http://cloudevents.io/schema.json";
 const subject = "subject.ext";
 const {
   BINARY_HEADERS_03,
-  HEADER_CONTENT_TYPE,
-  SPEC_V03
+  HEADER_CONTENT_TYPE
 } = require("../../../lib/bindings/http/constants.js");
 
 const ceContentType = "application/json";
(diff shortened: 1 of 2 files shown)


test/bindings/http/unmarshaller_0_3_tests.js
```diff
@@ -26,10 +26,11 @@ describe("HTTP Transport Binding Unmarshaller for CloudEvents v0.3", () => {
     const un = new Unmarshaller();
 
     // act and assert
-    return un.unmarshall(payload)
-      .then(() => { throw new Error("failed"); })
-      .catch((err) =>
-        expect(err.message).to.equal("payload is null or undefined"));
+    try {
```
danbev, line 29: Instead of try/catch perhaps this could use `chai`'s expect to check that this function throws:
```js
expect(() => un.unmarshall(null)).to.throw("payload is null or undefined");
```

lance: Really good suggestion, Dan. I'm implementing it now, and it in fact exposed a test that should have always been failing but was never noticed. [This test](https://github.com/cloudevents/sdk-javascript/pull/126/files#diff-7b7b330acdf1e0706f45a812d16cf558R73-R77) never throws and it should. But without using `expect(foo).to.throw()` the fact that it's not throwing is never caught. Anyhoo. I'm going to now figure out what the bug is and go from there.

PR #129, opened 2020-05-06
Inline Spec File

grant:
This PR inlines the spec file and deletes the `ext` folder. In the future, we can use a more sophisticated way of managing multiple spec versions.


grant:
> Inlining the spec file looks fine. 

Great.

> It seems like you have inadvertently pulled in a commit that has already landed 5b747ee. Can you remove this from the PR?

Yes, removed.

Issue #133, opened 2020-05-07
Unable to find parser given the encoding and headers provided


deewhyweb:
Using the latest code from master, I'm creating an event from Kafka which is coming into my app as:
**body:** "{\"specversion\":\"1.0\",\"id\":\"3\",\"type\":\"type\",\"source\":\"123\"}"

**headers:** {"host":"event-display.kafka.svc.cluster.local","user-agent":"Go-http-client/1.1","content-length":"59","accept-encoding":"gzip","ce-id":"partition:1/offset:23","ce-source":"/apis/v1/namespaces/kafka/kafkasources/kafka-source#knative-demo-topic","ce-specversion":"1.0","ce-subject":"partition:1#23","ce-time":"2020-05-07T14:16:30.245Z","ce-type":"dev.knative.kafka.event","forwarded":"for=10.131.0.72;proto=http","k-proxy-request":"activator","kourier-random-header":"558c917a-906d-11ea-96ab-0a580a830047","traceparent":"00-73df58d7c1383e17dabe515dfd68d05a-b00799e628f49cbe-00","x-envoy-expected-rq-timeout-ms":"600000","x-forwarded-for":"10.131.0.72, 10.128.2.99","x-forwarded-proto":"http","x-request-id":"d3649c1b-a968-40bf-a9da-3e853abc0c8b"}

The code I'm using is as follows:
```
const express = require('express');
const app = express();
const { HTTPReceiver, CloudEvent } = require('cloudevents-sdk');

const receiver = new HTTPReceiver();
app.use((req, res, next) => {
  let data = "";

  //req.setEncoding("utf8");
  req.on("data", function(chunk) {
      data += chunk;
  });
  req.on("end", function() {
      req.body = data;
      next();
  });
});
app.post('/', (req, res) => {
  console.log('Version 1.20');
  console.log('body:',JSON.stringify(req.body))
  console.log('headers:',JSON.stringify(req.headers))
  try {
    let myevent = receiver.accept(req.headers,req.body);
    console.log(JSON.stringify(myevent))
    res.status(201).send("Event Accepted");

  } catch(err) {
    // TODO deal with errors
    console.error(err);
    res.status(415)
          .header("Content-Type", "application/json")
          .send(JSON.stringify(err));
  }
});
const port = process.env.PORT || 8080;
app.listen(port, () => {
  console.log('Hello world listening on port', port);
});
```

I receive the following error:

> TypeError: Cannot read property 'parse' of undefined
[...]

PR #134, opened 2020-05-07
fix: ensure binary events can handle no content-type header

lance:
The fix provided in https://github.com/cloudevents/sdk-javascript/pull/118
only included tests for `receiver.check()`, and the change in that
case was to add the `application/json` content type to the cleansed
headers if no type was specified.

However, `receiver.parse()` did not receive the benefit of this change. It
calls `this.check()` but then sanitizes the original headers again, and the
missing content-type was not re-inserted into the newly sanitized headers.

This commit, modifies the code so that `receiver.check()` does not insert
the content-type, but does allow the validation check to pass if no
content-type header exists. When `receiver.parse()` is called, and the
headers are sanitized again - and this time used to look up parser implementation,
the default `application/json` content-is applied if no content-type header
exists.

I've also removed a redundant call to `receiver.check()` in receiver_binary_1.js
and simplified the usage of `Constants` in the test.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/133

### DIFF
--- a/lib/bindings/http/receiver_binary.js
+++ b/lib/bindings/http/receiver_binary.js
@@ -55,8 +55,9 @@ BinaryHTTPReceiver.prototype.check = function(payload, headers) {
 
   // Validation Level 1 - if content-type exists, be sure it's
   // an allowed type
-  if (sanityHeaders[HEADER_CONTENT_TYPE] && !this.allowedContentTypes
-    .includes(sanityHeaders[HEADER_CONTENT_TYPE])) {
+  const contentTypeHeader = sanityHeaders[HEADER_CONTENT_TYPE];
+  const noContentType = !this.allowedContentTypes.includes(contentTypeHeader);
+  if (contentTypeHeader && noContentType) {
     const err = new TypeError("invalid content type");
     err.errors = [sanityHeaders[HEADER_CONTENT_TYPE]];
     throw err;
--- a/lib/bindings/http/receiver_binary.js
+++ b/lib/bindings/http/receiver_binary.js
@@ -1,4 +1,5 @@
-const Constants = require("./constants.js");
+const { HEADER_CONTENT_TYPE, MIME_JSON, DEFAULT_SPEC_VERSION_HEADER } =
+  require("./constants.js");
 const Commons = require("./commons.js");
 const CloudEvent = require("../../cloudevent.js");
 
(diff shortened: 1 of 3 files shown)


lance:
@fabiojose @grant @helio-frota I was hoping to get this into the upcoming 2.0.0 release. Would you mind giving this a look today? 

I was aiming to release this week, but it's Friday and there are still some things that should be documented, so probably Monday for the release.

lib/bindings/http/receiver_binary.js
```diff
@@ -52,16 +53,12 @@ BinaryHTTPReceiver.prototype.check = function(payload, headers) {
...
-  if (!sanityHeaders[Constants.HEADER_CONTENT_TYPE]) {
-    sanityHeaders[Constants.HEADER_CONTENT_TYPE] = Constants.MIME_JSON;
-  }
-
-  // Validation Level 1
-  if (!this.allowedContentTypes
-    .includes(sanityHeaders[Constants.HEADER_CONTENT_TYPE])) {
+  // Validation Level 1 - if content-type exists, be sure it's
+  // an allowed type
+  if (sanityHeaders[HEADER_CONTENT_TYPE] && !this.allowedContentTypes
```
helio-frota, line 58: It would be nice to simplify that part to be a little better readable. But not required. 
The issue here is the wording + the `!` 
If ( foo is --> not_this_allowed_includes )

But at the same time I don't have a good suggestion for this... maybe this bellow ?
```
  const isNotIncludedHeader = !this.allowedContentTypes
    .includes(sanityHeaders[HEADER_CONTENT_TYPE]);

  if (sanityHeaders[HEADER_CONTENT_TYPE] && isNotIncludedHeader) {
```

test/bindings/http/promiscuous_receiver_test.js
```diff
@@ -67,17 +74,53 @@ describe("HTTP Transport Binding Receiver for CloudEvents", () => {
...
   });
+
+  describe("Kafka-Knative event source", () => {
+    const specversion = "1.0";
+    const id = "partition:1/offset:23";
+    const type = "dev.knative.kafka.event";
+    const source =
+      "/apis/v1/namespaces/kafka/kafkasources/kafka-source#knative-demo-topic";
+
+    it("Should be parseable", () => {
```
helio-frota, line 96: Minor typo `parsable` instead of `parseable`.

lib/bindings/http/receiver_binary.js
lance: Done. Thanks for the suggestion.

test/bindings/http/promiscuous_receiver_test.js
lance: Gotcha. Fixed.

PR #137, opened 2020-05-07
Introduce Typescript

grant:
This PR introduces TypeScript to this repo. This is the first step to fix the top-voted issue, #9.

- It includes a minimal setup for TypeScript.
- It does not introduce any features.
- It does not change any syntax you're used to, only renaming a file from `js` to `ts`

It does require running `npm run watch` to incrementally build or `npm run build` to build once. - For development, it's common to just run `npm run watch` in one tab.
- For publishing, it's common to run `npm run build` one time before publishing.

We can't really see benefits right now, but features such as autocompletion, refactoring, automatic type annotation would come for free.


package.json
```diff
@@ -92,15 +94,16 @@
...
     "eslint-config-standard": "^14.1.1",
     "eslint-plugin-import": "^2.20.2",
     "eslint-plugin-node": "^11.1.0",
-    "standardx": "^5.0.0",
-    "standard-version": "^7.1.0"
+    "mocha": "~7.1.1",
+    "nock": "~12.0.3",
+    "nyc": "~15.0.0",
+    "standard-version": "^7.1.0",
+    "standardx": "^5.0.0"
```
cwallsfdc, line 106: Is `typescript` needed as a `devDependencies`?

grant: Yes! Thanks. I've added it.

grant:
Closed in favor of #155.

Issue #124, opened 2020-05-05
Eliminate up-front spec version awareness in the emitter


lance:
Similar to https://github.com/cloudevents/sdk-javascript/issues/93, it should not be required for a user to specify the version when emitting events over HTTP. The default should be the most recent protocol version (1.0) and the user should be able to specify a previous supported version.

PR #141, opened 2020-05-11
feat!: expose a version agnostic event emitter

lance:
This is a breaking change.

This commit exposes an HTTP based event emitter that simplifes the API.
To use it, simply import the SDK and start emitting. The default spec
version is 1.0, but you can use 0.3 by supplying that to the constructor.

By default, CloudEvents are emitted in binary mode, but this can be changed
by providing the "structured" parameter to the `emit()` function.

This commit also eliminates the version specific emitters and receivers
from the `v1` and `v03` exports, and eliminates the explicit usage of
versioned emitters from `lib/bindings/http`.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/124 

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,25 +62,52 @@ console.log(receivedEvent.format());
 
 #### Emitting Events
 
-Currently, to emit events, you'll need to decide whether the event is in
+To emit events, you'll need to decide whether the event should be sent in
 binary or structured format, and determine what version of the CloudEvents
 specification you want to send the event as.
 
+By default, the `HTTPEmitter` will emit events over HTTP POST using the
+1.0 specification, in binary mode. You can emit 0.3 events by providing
+the specication version in the constructor to `HTTPEmitter`. To send
+structured events, add that string as a parameter to `emitter.sent()`.
+
 ```js
-const { CloudEvent } = require("cloudevents-sdk");
-const { StructuredHTTPEmitter } = require("cloudevents-sdk/v1");
+const { CloudEvent, HTTPEmitter } = require("cloudevents-sdk");
 
-const myevent = new CloudEvent()
-  .type("com.github.pull.create")
-  .source("urn:event:from:myapi/resource/123");
+// Without any parameters, this creates a v1 emitter
+const v1Emitter = new HTTPEmitter();
+const event = new CloudEvent()
+  .type(type)
+  .source(source)
+  .time(new Date())
+  .data(data)
 
-const emitter = new StructuredHTTPEmitter({
-  method: "POST",
-  url   : "https://myserver.com"
-});
+// By default, the emitter will send binary events
+v1Emitter.send({ url: "https://cloudevents.io/example" }, event)
+  .then((response) => {
+    // handle the response
+  })
+  .catch(console.error);
 
-// Emit the event
-emitter.emit(myevent)
+// To send a structured event, just add that as a parameter
+v1Emitter.send({ url: "https://cloudevents.io/example" }, event, "structured")
+  .then((response) => {
+    // handle the response
+  })
+  .catch(console.error);
+
+// Sending a v0.3 event works the same, just let the emitter know when
+// you create it that you are working with the 0.3 spec
+const v03Emitter = new HTTPEmitter("0.3");
+
+// Again, the default is to send binary events
+// To send a structured event, add "structured" as a final parameter
+v3Emitter.send({ url: "https://cloudevents.io/example" }, event)
+  .then((response) => {
+    // handle the response
+  })
+  .catch(console.error);
+
 ```
 
 ## Supported specification features
--- a/README.md
+++ b/README.md
@@ -74,8 +74,10 @@ structured events, add that string as a parameter to `emitter.sent()`.
 ```js
 const { CloudEvent, HTTPEmitter } = require("cloudevents-sdk");
 
-// Without any parameters, this creates a v1 emitter
-const v1Emitter = new HTTPEmitter();
+// With only an endpoint URL, this creates a v1 emitter
+const v1Emitter = new HTTPEmitter({
+  url: "https://cloudevents.io/example"
+});
 const event = new CloudEvent()
   .type(type)
   .source(source)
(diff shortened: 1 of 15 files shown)


lib/bindings/http/emitter_binary.js
```diff
@@ -1,49 +1,85 @@
...
+ */
+class BinaryHTTPEmitter {
+  /**
+   *
+   * @param {string} version - the CloudEvent HTTP specification version.
+   * Default: 1.0
+   */
+  constructor(version) {
+    if (version === SPEC_V1) {
+      this.headerByGetter = require("./emitter_binary_1");
```
danbev, line 32: Perhaps add the `.js` suffix/extension here for consistency?

lib/bindings/http/emitter_binary.js
```diff
@@ -1,49 +1,85 @@
...
+   */
+  async emit(options, cloudevent) {
+    const config = { ...options, ...defaults };
+    const headers = config[HEADERS];
+
+    Object.keys(this.headerByGetter)
+      .filter((getter) => cloudevent[getter]())
+      .forEach((getter) => {
+        const header = this.headerByGetter[getter];
+        headers[header.name] =
```
danbev, line 58: Could this be on a single line, I think it helps make it easier to read.

test/sdk_test.js
```diff
@@ -1,69 +1,18 @@
...
-      expect(v03).to.have.property("BinaryHTTPReceiver");
-    });
-
-    it("should exports 'HTTPUnmarshaller'", () => {
-      expect(v03).to.have.property("HTTPUnmarshaller");
-    });
-
-    it("should exports 'event'", () => {
-      expect(v03).to.have.property("event");
+      expect(new CloudEvent(SpecV03).spec.payload.specversion).to.equal("0.3");
```
danbev, line 9: Nit: You can use the constant `SPEC_V03` here.

test/sdk_test.js
```diff
@@ -1,69 +1,18 @@
...
-      expect(v1).to.have.property("BinaryHTTPEmitter");
-    });
-
-    it("should exports 'BinaryHTTPReceiver'", () => {
-      expect(v1).to.have.property("BinaryHTTPReceiver");
-    });
-
-    it("should exports 'event'", () => {
-      expect(v1).to.have.property("event");
+      expect(new CloudEvent(SpecV1).spec.payload.specversion).to.equal("1.0");
```
danbev, line 15: Nit: same as above, perhaps use `SPEC_V1`.

lance:
@danbev thanks for the review - I think I was a bit early in submitting this. I want to add more tests and docs. Marking as WIP for now.

lib/bindings/http/http_emitter.js
```diff
@@ -0,0 +1,56 @@
...
+   * @param {Object} options The configuration options for this event. Options
+   * provided will be passed along to Node.js `http.request()`.
+   * https://nodejs.org/api/http.html#http_http_request_options_callback
+   * @param {URL} options.url The HTTP/S url that should receive this event
+   * @param {CloudEvent} event the CloudEvent to be sent
+   * @param {string} [mode] the message mode for sending this event.
+   * Possible values are "binary" and "structured". Default: structured
+   * @returns {Promise} Promise with an eventual response from the receiver
+   */
+  send(options, event, mode = "binary") {
```
lance, line 45: @danbev any thoughts on this as a function signature? I'm not sure whether it makes sense to have an emitter always emit in binary or structured, or instead have the `mode` parameter in the ctor and once constructed it only emits in a single mode. Also... I don't love the whole `Constants` construct, but since it's everywhere, do you think we should have constants for `"binary"` and `"structured"`? 

Also... the first parameter is an `options` object, but the only required property is `url`. I had considered making the URL a string parameter, but we still need to accept an options object because this allows callers to pass things like `HttpAgent` here, which eventually makes its way to `http.request()`.

lance:
OK - I think I've covered my bases here. Tests and docs added. PTAL. 

@danbev @helio-frota if you wouldn't mind another review, I'd appreciate it.

lib/bindings/http/http_emitter.js
danbev: > I'm not sure whether it makes sense to have an emitter always emit in binary or structured

My first thought was that it makes sense but thinking about this a little more I'm not so sure. For example, If I'm using this HttpEmitter then I know the destination of these events, the url we are posting to. If this destination is some kind of message broker I might be able to chose the best format for it (even though if can/must handle both it might always normalize into a specific content format). In that case I'm thinking just using one mode would be enough. Is my reasoning sane here?

 >I don't love the whole Constants construct, but since it's everywhere, do you think we should have constants for "binary" and "structured"?

Yeah, I think we should and then later perhaps look at a better solution for the constants.

.eslintrc
```diff
@@ -13,7 +13,7 @@
     "arrow-parens": ["error", "always"],
     "arrow-body-style": ["error", "as-needed"],
     "prefer-template": "error",
-    "max-len": ["warn", { "code": 80 }],
+    "max-len": ["warn", { "code": 120 }],
```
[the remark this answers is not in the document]
helio-frota: I'm ok with 120 :+1:

lib/bindings/http/http_emitter.js
lance: Thanks for the feedback @danbev. I've started [a thread](https://cloud-native.slack.com/archives/CCXF3CBL1/p1589377588205000) in the CE slack looking for feedback/opinions. If you have any further thoughts on this, we can hash it out there.

Will add "binary" and "structured" to the constants module. Will probably pull these up to be more accessible via the top level import. E.g.

lance:
@helio-frota @danbev - I've [modified the API](https://github.com/cloudevents/sdk-javascript/pull/141/files#diff-04c6e90faac2675aa89e2176d2eec7d8R77-R117) a little bit. If you have a chance, PTAL.

README.md
```diff
@@ -62,25 +62,52 @@ console.log(receivedEvent.format());
...
 
-```js
-const { CloudEvent } = require("cloudevents-sdk");
-const { StructuredHTTPEmitter } = require("cloudevents-sdk/v1");
-
-const myevent = new CloudEvent()
-  .type("com.github.pull.create")
-  .source("urn:event:from:myapi/resource/123");
+By default, the `HTTPEmitter` will emit events over HTTP POST using the
+1.0 specification, in binary mode. You can emit 0.3 events by providing
```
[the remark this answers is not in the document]
lance: @erikerikson that's a valid point. I would like to avoid having that hold up the PR, so I've added a new issue to address it #160. Work for you?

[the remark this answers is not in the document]
erikerikson: Works for me

PR #147, opened 2020-05-12
chore: adds files section in package.json

helio-frota:


helio-frota:
Commit updated to remove redundancy with `index.js` already declared on `main` section.

PR #158, opened 2020-05-13
lib: removes unused array

helio-frota:
Signed-off-by: Helio Frota <00hf11@gmail.com>


helio-frota:
@lance probably yes. The same array is in use on 0.3 tests

Issue #162, opened 2020-05-14
Add JSDoc to BinaryHTTPReceiver and StructuredHTTPReceiver


lance:
This also may involve converting these to ES6 classes.

PR #165, opened 2020-05-15
lib!: refactor HTTP bindings and specifications

lance:
This is a breaking change.

This commit makes a number of changes to the HTTP bindings code in an attempt
to simplify its usage and implementation. From a very high level, this inverts
the existing dependencies.

As an example, consider `lib/bindings/http/receiver_structured_1.js`.
https://github.com/cloudevents/sdk-javascript/blob/v1.0.0/lib/bindings/http/receiver_structured_0_3.js

This class instantiates `lib/bindings/http/receiver_structured.js` and delegates
its function invokations to it. This had the effect of requiring a user to know what
event versions they would be receiving. And for me personally was a little confusing
as a maintainer.

The change introduced here reverses that logic, so that the version agnostic receiver
is what the user instantiates. It instantiates the approrpiate version of a specific
receiever and delegates to it - reversing the dependencies.

I've also moved all of the top level directories related to HTTP versions into
`lib/bindings/http/v1` and `lib/bindings/http/v03` and generally done some rearranging
to make the repository structure cleaner and more organized.

As a side effect:

Fixes: https://github.com/cloudevents/sdk-javascript/issues/162


docs/bindings_http_receiver_binary.js.html
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Parses an incoming HTTP request, converting it to a {CloudEvent}
+   * instance if it conforms to the Cloud Event specification for this receiver.
+   *
+   * @throws {ValidationError} of the event does not conform to the spec
+   * @param {Object} payload the HTTP request body
+   * @param {Object} headers the HTTP request headers
+   * @returns {CloudEvent} aninstance of CloudEvent representing the incoming request
```
helio-frota, line 139: Missing space `aninstance`

grant:
Sorry I didn't get to review this in time.

It seems like this PR has a lot of autogenerated HTML that will be in the npm package that isn't useful for consumers. Is that on purpose?

lance:
Hey @grant. With @helio-frota's PR #147 last week, the package.json has been modified so that it will only include the source code - no docs or tests or examples. The `npm run generate-docs` generates all of these `.html` files and they are published here: https://cloudevents.github.io/sdk-javascript/.

lance:
> Sorry I didn't get to review this in time.

:wink: this is why I was proposing a grace period - after a PR is reviewed, but before it can be landed.

Issue #148, opened 2020-05-12
CloudEvent should generate a date by default


lance:
See: https://github.com/knative/docs/pull/2446/files#diff-de16006ea59029f2429d3228fed269c6R26

lance:
From the [spec](https://github.com/cloudevents/spec/blob/v1.0/spec.md#time), this should conform to [RFC 3339](https://tools.ietf.org/html/rfc3339)

Issue #135, opened 2020-05-07
HTTP receiver should not expose bookkeeping functions


lance:
In [receiver_binary.js](https://github.com/cloudevents/sdk-javascript/blob/master/lib/bindings/http/receiver_binary.js#L44-L79) there is a `check()` function which is used to validate the input. It is currently defined on `BinaryHTTPReceiver.prototype` but I don't think it should be exposed at all. It's the SDK's job to check input and validate it, not the user of the SDK, so why expose it? I suspect it's useful for testing. In fact, that's how the [tests](https://github.com/cloudevents/sdk-javascript/blob/master/test/bindings/http/receiver_binary_1_tests.js#L14) are written at the moment.  The same exists for the [structured receiver](https://github.com/cloudevents/sdk-javascript/blob/master/lib/bindings/http/receiver_structured.js#L36-L50).

Since the `parse()` function already calls `check()`, I think it's reasonable to remove this function from the `prototype` and make it internal to the module.

Are there any objections to this?

Issue #88, opened 2020-04-29
Consider refactoring Cloudevent.format()


lance:
Based on some discussion in [this pull request](https://github.com/cloudevents/sdk-javascript/pull/87#issuecomment-621412733) there are some improvements that could be made to the [`Cloudevent.format()` function](https://github.com/cloudevents/sdk-javascript/blob/master/lib/cloudevent.js#L20-L29).

* Consider optimizing so that JSON.parse doesn't require several levels of indirection to get to.
* Expose a way to retrieve the data in multiple formats
* Consider moving the sanity checks into the `spec` instance and not in `Cloudevent` itself.

PR #172, opened 2020-05-20
lib!: change CloudEvent to use direct object notation and get/set properties

lance:
This commit makes a substantial change to the API, changing the CloudEvent class
to accept properties as an object in the constructor. For example:

```js
const CloudEvent = require('cloudevents-sdk');

// all event properties except extensions may be set in the constructor
const event = new CloudEvent({
  source: 'http://my.event.source',
  type: 'test-event-type'
});

// get and set all properties standard property notation
console.log(event.time); // the event timestamp
event.subject = 'my event subject';
```

This change includes a new test file to explicitly test this interface,
as well as new documentation.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/65
Fixes: https://github.com/cloudevents/sdk-javascript/issues/148
Fixes: https://github.com/cloudevents/sdk-javascript/issues/135
Fixes: https://github.com/cloudevents/sdk-javascript/issues/88


lance:
I realize there are a lot of changes in the HTML, however, the HTML is auto-generated, so usually I don't think it's necessary to spend time reviewing it. It's just reflecting what's in the code. Typically, I will just click the "Viewed" checkbox on the PR so they don't get in the way. Since our docs are configured to be published at https://cloudevents.github.io/sdk-javascript/ we need to have them on `master` instead of another branch. :man_shrugging:

grant:
I would really like no autogenerated files in source control. We can think of some other way to get the docs besides this. Not sure if others agree.

lance:
I think it's pretty common to use the `docs/` dir as a place for a project's website. That's what [sdk-go does](https://cloudevents.github.io/sdk-go/), for example. Of course, they have godoc.org to publish their API which we don't have - and which is why our generated documentation lives in `docs/`. I don't think these will be changing regularly - only when there are big API changes like this.

lance:
Since it's such a big change, if you just want to see how the API is used, I added some tests that just exercise the new interface. https://github.com/cloudevents/sdk-javascript/pull/172/files#diff-dd013ce6952212675e25f44a069d648a

lib/bindings/http/validation/structured.js
```diff
@@ -14,8 +14,7 @@ function check(payload, headers, receiver) {
   const sanityHeaders = sanityAndClone(headers);
 
   // Validation Level 1
-  if (!receiver.allowedContentTypes
-    .includes(sanityHeaders[HEADER_CONTENT_TYPE])) {
+  if (!receiver.allowedContentTypes.includes(sanityHeaders[HEADER_CONTENT_TYPE])) {
```
helio-frota, line 17: Not a blocker, but I remember something like that in other file (?) : ) 

if ( not_receiver_allowed_includes foo )

lib/cloudevent.js
```diff
@@ -1,232 +1,314 @@
...
+const Spec03 = require("./bindings/http/v03/spec_0_3.js");
 const Formatter = require("./formats/json/formatter.js");
 
+const { SPEC_V1, SPEC_V03 } = require("./bindings/http/constants.js");
+const { isBinary } = require("./bindings/http/validation/fun.js");
+
 /**
- * An instance of a CloudEvent.
+ * An CloudEvent describes event data in common formats to provide
+ * interopability across services, platforms and systems.
```
helio-frota, line 10: typo `interopability`

helio-frota:
regarding some core business changes that may exist here and there I have no much idea, other than that it looks good the move from prototype to ES6 classes. LGTM after the minor changes :+1:

lib/bindings/http/validation/structured.js
helio-frota: Sorry missing the context here, I remember that we saw something like that in other file.
and you extracted some variables to improve the readability.

helio-frota: @lance I found it https://github.com/cloudevents/sdk-javascript/pull/134#discussion_r422346076

lance: Ahh gotcha. The difference here is that this is for a structured receiver. The PR you point to is modifying a binary event receiver where an event is sent without any content-type header. For structured receivers, [the content-type must be there](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#321-http-content-type).

PR #181, opened 2020-05-21
chore: add GitHub action for CI on master and prs

lance:


lance:
It took less than 20 seconds for each of Node 10 and 12. I think this might be a better solution that Travis currently.

Issue #188, opened 2020-05-27
Generated type definitions don't declare optional parameters 


lance:
When using the new direct object style API to create a `CloudEvent` in a TypeScript project, the `CloudEvent` constructor requires all properties for the event to be provided.

This is what the constructor signature looks like in the generated type definition.

```js
    constructor({ id, source, type, dataContentType, time, subject, dataSchema, schemaURL, dataContentEncoding, data, specversion }?: {
        source: string; 
        type: string;
        id: string; 
        time: string;
        subject: string;
        dataContentType: string;
        dataSchema: string;
        schemaURL: string;
        dataContentEncoding: string;
        specversion: string;
        data: any;
    });
```

There are two things wrong with this. First of all, it shows the single properties parameter as being optional, and it indicates that each of the individual fields of this properties parameter is required. This does not happen in a pure JavaScript project because the autocompletion feature in VSCode for JS projects that also have type definitions prefers the `.js` version over the `.d.ts` version. In a TypeScript project, however, the VSCode autocompleter only considers the `.d.ts` file.

Until a fix is in place, there may be a work around by adding some TypeScript configuration option. But I'm not sure what that is - don't have enough experience with it yet.

grant:
This seems like a blocker for projects that use this library using TypeScript and require optional types.
A solution would be to declare a CloudEvent interface with optional fields, but that obviously uses TypeScript which we can't have.
Curious to see the solution.

lance:
@grant I have been reading and playing around with this for the last hour and was going to come back here and ask if perhaps an interface would fix this. And I find this.

> but that obviously uses TypeScript which we can't have.
> Curious to see the solution.

For crying out loud, please can't you just be civil for once? 

Yes, let's please find the right solution. I have an inelegant quick fix here: https://github.com/cloudevents/sdk-javascript/pull/189.

I am surprised that you let us push a 2.0.0 release without noticing this. Did you not try to use the SDK in a TypeScript project before approving https://github.com/cloudevents/sdk-javascript/pull/155?

lance:
I assume an interface would address #184 as well.

grant:
Yes, sorry, that was a bit impolite, @lance. I've edited the comment.
Still genuinely interested in seeing the technological though.

Issue #184, opened 2020-05-22
CloudEvent JSON Autocompletion without Class Wrapper


grant:
As a developer, I want to be able to have autocompletion for objects without having a wrapper class such as `CloudEvent`. 

JavaScript objects are normal in Node development, and are preferred over classes.

![image](https://user-images.githubusercontent.com/744973/82617377-c3baa980-9b84-11ea-9ea0-2bb12e3b4f89.png)

**EXPECTED:**

```ts
import {CloudEvent} from 'cloudevents/v1';
const ce = {
    "specversion" : "1.0",
    "type" : "com.github.pull.create",
    "source" : "https://github.com/cloudevents/spec/pull",
    "subject" : "123",
    "id" : "A234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "datacontenttype" : "text/xml",
    "data" : "<much wow=\"xml\"/>"
}: CloudEvent;
ce.
```

**ACTUAL:**

```ts
import {CloudEvent} from 'cloudevents/v1';
const ce = new CloudEvent({
    "specversion" : "1.0",
    "type" : "com.github.pull.create",
    "source" : "https://github.com/cloudevents/spec/pull",
    "subject" : "123",
    "id" : "A234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "datacontenttype" : "text/xml",
    "data" : "<much wow=\"xml\"/>"
});
ce.
```

> **Note:** The `: CloudEvent` can be dropped off from user code and be inferred with normal usage if the developer doesn't like TypeScript.

### Advantages

- Minus one level of indirection (no required Classes/Objects).
- Better devX: Creating `new CloudEvent()`, `new CloudEvent()`, over and over is tedious and one more level of indirection. It's much easier to use JSON. I think most Node developers would agree.
- Support common JS features: You can't combine 2 CloudEvent JSON payloads easily, but you can with 2 JavaScript objects (`{ ...ce1, ...ce2}`).
- Performance: JS Class wrappers for JSON creates unnecessary constructors to be called. Simply not requiring a class can help.
- Use raw JSON over JS Classes. (It's easier to copy/paste a CloudEvent into source code without wrapping it.

### Disadvantages

[...]

PR #189, opened 2020-05-27
fix: add two type definitions to SCM and fix their parameters

lance:
It seems that `tsc` doesn't recognize when a class or function has a single
`options` parameter that is required but properties within it are not.
This corrects the output from `tsc` type definition generation and adds
the two files affected by this to repo.

This might not be the best way to go, but it is a quick fix that we should at least discuss.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/188


grant:
This is certainly one solution. Thanks for trying to get the types in here!

However I think a more elegant solution would just be to not create new files (`.d.ts`) that are hand-modified, and to just change the files to normal files (`.ts` without the (`.d`). By excluding certain `.js` files, it looks like we are manually creating these type definitions, which is not sustainable or conventional.

I think we can all agree that we shouldn't include generated files in source control.
Happy to hear thoughts around this.

lance:
Thanks for the feedback. Can you put together a PR that illustrates what you are talking about and perhaps uses an interface to address it?

grant:
I'd imagine something like this:

```js
constructor(cloudEvent: CloudEventV1 = SPEC_V1) {
 ...
}
```

In a `cloudevent.ts` file:
https://github.com/cloudevents/sdk-javascript/blob/master/lib/cloudevent.js#L29

---

The source for `CloudEventV1` would be something like this:
https://github.com/cloudevents/sdk-javascript/issues/184#issuecomment-632408537

Unfortunately I don't think we can add optional types without changing the extension or manually creating types as suggested in this PR.

I'd rather rename the file extension as generated-then-hand-modified files are hard to keep up-to-date.

lance:
@grant if you can put together a PR for that I would appreciate it. I think that will be easiest for me to understand how it all fits together, and to experiment with it from both a pure JS project and a pure TS project.

grant:
> @grant if you can put together a PR for that I would appreciate it. I think that will be easiest for me to understand how it all fits together, and to experiment with it from both a pure JS project and a pure TS project.

I'm happy to put a PR together if it will be accepted. However, I don't know if you understand what I'm suggesting. This would be my 3rd attempt at introducing `.ts` to CloudEvents, and I have moved on to other languages until we can solve this.

From a technical standpoint, I would like to ensure we understand:
- We're OK with these filename change:
  - `lib/cloudevent.js` -> `"lib/cloudevent.ts"
  - `lib/bindings/http/http_emitter.js` -> `lib/bindings/http/http_emitter.ts`
- We're OK with using a uniquely TypeScript feature, `interface` to support optional types in objects. Example: https://github.com/cloudevents/sdk-javascript/issues/184#issuecomment-632408537
- We're willing to use continually use `tsc --watch` to continuously build `.ts` files into a `.js` files.

If we are not happy with those understandings, I can't see a technical way to solve this problem and can't contribute. if we're OK with that, I am free Friday to collaborate and put a PR together.

lance:
Yes I think this could be ok. I am not too concerned about the filename changes. I don’t love the continuous `tsc` but I can live with it. What I want to better understand is how an `interface` fits in. What is the filename? How does it affect (if at all) the way a pure JS developer interacts with the SDK? Do any of the other type definitions need to know about it? Etc.

grant:
Hey Lance, I'll try to answer some of the questions.
TL;DR; Javascript users will get the same experience. TS developers and SDK developers will get better autocompletion and less code duplication.

> Yes I think this could be ok. I am not too concerned about the filename changes. I don’t love the continuous `tsc` but I can live with it.

OK. Yeah it's sorta bothersome in the beginning, but it's really lightweight and provides a pretty good devX imo.

> What I want to better understand is how an `interface` fits in. What is the filename?

So, right now we are generating `.ts` files with `tsc` on `js` files. The generated files are named `*.d.ts`, still `.ts` files. Tsc infers the types and generates declarations. Ctrl+click to see the definition of the types. OK autocompletion but not great. I know it's painful the first time you look at a `.ts` file with interfaces, but it's worth understanding [interfaces](https://www.typescriptlang.org/docs/handbook/interfaces.html).

A TypeScript `interface` provides structure to a plain old JavaScript object. It's used by editors like VS Code and others, even if you're using plain JavaScript. Some npm modules.

It affects the SDK code writing `{ id, source, type, dataContentType, time, subject, dataSchema, schemaURL, dataContentEncoding, data, specversion }` multiple times (no wrapper class).

When we publish the npm module, we publish our code which has a `package.json` file. This file can have a `types` property. This `types` property is used by editors to view the types.

> How does it affect (if at all) the way a pure JS developer interacts with the SDK? Do any of the other type definitions need to know about it? Etc.

It doesn't affect the pure JS end developer. Pure JS developers do however often use editors like VS Code that provides inference via TS. TypeScript does not attempt to force the JS world to type TypeScript, it's strictly optional because it is a strict language superset.

> I meant what I said when I said #155 could be the start of a gradual movement towards TS. And I admit that this comes from ignorance on my part. But I don’t think that I would have been able to get a 2.0 release out as quickly as I did if I had to also be learning the machinations of typescript in the meantime.

Great. It's fine to have everyone learn and contribute to the project with their own perspectives and requirements.

> I’m happy to start making that journey right now.

Great to hear.

lance:
Closing in favor of #194

PR #194, opened 2020-05-28
fix: introduce CloudEventV1 and CloudEventV03 interfaces

lance:
This commit introduces `CloudEventV1` and `CloudEventV03` TypeScript interfaces. Additionally, it changes all of the user facing API to be `.ts` instead of `.js` files. Those are:

* `lib/cloudevent.ts`
* `lib/bindings/http/http_receiver.ts`
* `lib/bindings/http/http_emitter.ts`
* `lib/bindings/http/constants.ts`

### Usage in JavaScript
The pure JS API has not changed. You still create a `CloudEvent` like this.

```js
const { CloudEvent } = require('cloudevents-sdk');

const ce = new CloudEvent({
  source: 'a source',
  type: 'type'
});
console.log(ce);
```
```console
❯ node index.js
CloudEvent {
  spec:
   Spec1 {
     payload:
      { specversion: '1.0',
        id: 'b3b875c3-f7f0-410f-8aca-6ddcc61625b4',
        time: '2020-05-28T20:51:07.033Z',
        source: 'a source',
        type: 'type',
        datacontenttype: undefined,
        data: undefined,
        subject: undefined } },
  formatter: JSONFormatter {},
  extensions: [] }
```

### Usage in TypeScript
This change resolves the auto-completion issue with raw JavaScript objects through the use of the new interfaces.

Example:
```js
import { CloudEvent } from 'cloudevents-sdk';
import { Event } from 'cloudevents-sdk/lib/cloudevent';
import { CloudEventV1 } from 'cloudevents-sdk/lib/v1';
import { CloudEventV1Attributes } from 'cloudevents-sdk/lib/v1';

const ce: CloudEventV1 = {
  specversion: '1.0',
  source: '/some/source',
  type: 'example',
  id: '1234',
};
const event = new CloudEvent(ce);

const ce2: CloudEventV1Attributes = {
  source: '/',
  type: 'example'
};
const event2 = new CloudEvent(ce2);

const event3 = new CloudEvent({
  source: '/',
  type: 'example'
});

const e: Event = {
  source: '/',
  type: 'example'
};
const ce4 = new CloudEvent(e);
```

### Documentation
The existing documentation in `./docs` was removed and should replaced with generated
HTML from tsdocs, pending some other method of publishing API documentation. To keep
this commit relatively small and focused on the code, that will come as a separate docs-only PR.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/184
Fixes: https://github.com/cloudevents/sdk-javascript/issues/188


tsconfig.json
```diff
@@ -3,14 +3,25 @@
...
     "esModuleInterop": true,                  /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
     "forceConsistentCasingInFileNames": true,  /* Disallow inconsistently-cased references to the same file. */
+    "outDir": "./",
+    "declaration": true
   },
+  "compileOnSave": true,
   "files": [
-    "index.js"
-  ]
+    "src/index.ts",
```
grant, line 17: Let's glob:
`src/**/*.ts`

lance:
@grant what do you think about the tradeoffs re: auto completion? As this PR stands now, the only way to create a `CloudEvent` is with `id` and `specversion` required. Two options to eliminate the requirement that I see. Change the interface or accept `any`.

lance:
@grant please re-review with these new changes: https://github.com/cloudevents/sdk-javascript/pull/194/commits/26095e77fcc5fc2a33f83b99bc7c868bb2e0cff9

src/lib/cloudevent.ts
```diff
@@ -8,6 +8,8 @@ import { isBinary } from "./bindings/http/validation/fun.js";
 
 const { SPEC_V1, SPEC_V03 } = require("./bindings/http/constants.js");
 
+export type Event = CloudEventV1 | CloudEventV1Attributes | CloudEventV03 | CloudEventV03Attributes
```
grant, line 11: `type CloudEvent`

src/lib/bindings/http/http_emitter.ts
```diff
@@ -78,17 +85,20 @@ class HTTPEmitter {
...
+function headers(event: CloudEvent, version = SPEC_V1) {
   const headers = {};
   let headerMap;
   if (version === SPEC_V1) {
     headerMap = EmitterV1;
   } else if (version === SPEC_V03) {
     headerMap = EmitterV03;
   }
-  headerMap.forEach((parser, getterName) => {
+  headerMap.forEach((parser: any, getterName: string) => {
```
grant, line 96: What is the type of `parser`? We probably want to make the `Base64Parser` and `JSONParser` extend the same parser.

We probably don't need classes though, since these parsers should really be methods and simplified.

grant:
Nice job applying the interface union @lance.

src/lib/bindings/http/http_emitter.ts
lance: I would prefer to open a new issue/pr for this. There are a bunch of changes that still could be done to the codebase to improve it overall - nits like `no-any` as you mention above as well as some of these bigger style/architecture concerns.

lance:
OK - this is landing! I will cut a 2.0.1 release either over the weekend or on Monday.

Issue #204, opened 2020-06-03
Simplify README


grant:
- Use forward English: Subject Verb Object rather than Object Verb Subject
- Be terse without losing information
- Use modern Node (no `--save`)

PR #205, opened 2020-06-03
Simplify README.md Header

grant:
Simplifies the README header. Let's make this repo really simple to use.

Fixes: #204

- We have a license file, GitHub/npm makes it obvious with various tooling.
- We're already on major version 2+, it's common for API versions and client versions to differ, no need to mention it.


README.md
```diff
@@ -5,37 +5,24 @@
...
+_Note:_ Supports CloudEvent versions 0.3, 1.0
 
-### Installation
+## Installation
 
 The CloudEvents SDK requires a current LTS version of Node.js. At the moment
 those are Node.js 10.x and Node.js 12.x. To install in your Node.js project:
 
 ```console
-npm install --save cloudevents-sdk
```
lance, line 38: It's pretty common to have either `-s` or `--save` in instructions like this. If you're adding it to your project, you're going to want it added to `package.json`, so why remove this?

grant: It's not required and is the default as of npm 5: https://stackoverflow.com/a/19578808/1233286

It has no functionality. It's common for older Node projects.

PR #210, opened 2020-06-05
feat: ts formatter

grant:
Uses `any` for now as many types are a bit whack with the hybrid ts/js repo.

### DIFF
--- a/src/lib/formats/json/formatter.ts
+++ b/src/lib/formats/json/formatter.ts
@@ -3,13 +3,13 @@ class JSONFormatter {
    * Every internal data structure is JSON by nature, so
    * no transformation is required
    */
-  format(payload) {
+  format(payload: any) {
     return payload;
   }
 
-  toString(payload) {
+  toString(payload: any) {
     return JSON.stringify(payload);
   }
 }
 
-module.exports = JSONFormatter;
+export default JSONFormatter;
--- a/src/lib/formats/json/parser.ts
+++ b/src/lib/formats/json/parser.ts
@@ -2,16 +2,17 @@ const {
   isString,
   isDefinedOrThrow,
   isStringOrObjectOrThrow
-} = require("../../bindings/http/validation/fun.js");
+} = require("../../bindings/http/validation/fun");
 const ValidationError = require("../../bindings/http/validation/validation_error.js");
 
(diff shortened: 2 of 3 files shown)


src/lib/formats/json/parser.ts
```diff
@@ -8,10 +8,11 @@ const ValidationError = require("../../bindings/http/validation/validation_error
 const invalidPayloadTypeError = new ValidationError("invalid payload type, allowed are: string or object");
 const nullOrUndefinedPayload = new ValidationError("null or undefined payload");
 
-const asJSON = (v) => (isString(v) ? JSON.parse(v) : v);
+const asJSON = (v: any) => (isString(v) ? JSON.parse(v) : v);
```
lance, line 11: I believe this could be 
```suggestion
const asJSON = (v: string|object) => (isString(v) ? JSON.parse(v) : v);
```
Since these are the only two valid options.

src/lib/formats/json/formatter.ts
```diff
@@ -3,13 +3,13 @@ class JSONFormatter {
    * Every internal data structure is JSON by nature, so
    * no transformation is required
    */
-  format(payload) {
+  format(payload: any) {
```
lance, line 6: Can probably be 
```suggestion
  format(payload: string) {
```

src/lib/formats/json/parser.ts
```diff
@@ -8,10 +8,11 @@ const ValidationError = require("../../bindings/http/validation/validation_error
...
 
 class JSONParser {
-  constructor(decorator) {
+  decorator: any
+  constructor(decorator: any) {
```
lance, line 15: The only decorator used at the moment is the `Base64Parser`. Could this be `{}`?

src/lib/formats/json/formatter.ts
lance: Obviously that didn't work

src/lib/formats/json/parser.ts
lance: I've successfully tested changing this to `Base64Parser` locally.

grant: Fixed. Thanks.

grant:
DCO instructions are incorrect. :(

PR #214, opened 2020-06-08
feat: pass extension into the constructor.

### LINKED ISSUE #209: CloudEvent constructor should accept inline extensions
There is no way to create a `CloudEvent` instance using the constructor and have the extensions initialized there. For example, this is not valid, but should work.

```js
const ce = new CloudEvent({
  type: 'ex',
  source: '/path',
  extensions: {
    extension1: 'some data'
  }
});
```

lholmquist:
* This allows someone to pass an extension/extensions into the CloudEvent contructor when creating a CloudEvent.

fixes #209

### DIFF
--- a/test-ts/cloud_event_test.ts
+++ b/test-ts/cloud_event_test.ts
@@ -119,6 +119,16 @@ describe("A 1.0 CloudEvent", () => {
     expect(Object.keys(ce.extensions).length).to.equal(0);
   });
 
+  it("can be constructed with extensions", () => {
+    const extensions = {
+      "extension-key": "extension-value"
+    };
+    const ce = new CloudEvent({
+      extensions, ...fixture
+    });
+    expect(Object.keys(ce.extensions).length).to.equal(1);
+  });
+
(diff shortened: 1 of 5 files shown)


test-ts/cloud_event_test.ts
```diff
@@ -119,6 +119,16 @@ describe("A 1.0 CloudEvent", () => {
...
   });
 
+  it("can be constructed with extensions", () => {
+    const extensions = {
+      "extension-key": "extension-value"
+    };
+    const ce = new CloudEvent({
+      extensions, ...fixture
+    });
+    expect(Object.keys(ce.extensions).length).to.equal(1);
```
lance, line 129: You may want to also check the value in a second `expect()`.

lholmquist: yea,  let me add that

lholmquist: Added another test in the latest commit

lholmquist:
Added an Extension interface as well as tests

Issue #223, opened 2020-06-12
Passing custom headers through to http.request()


c-pius:
When using HttpEmitter.send() method, how can I pass additional header arguments (e.g. meant for our infrastructure) that are passed through to the http.request() call? 

I tried to use
```javascript
const emitter = new HTTPEmitter({
    url: "https://foo.bar"
})

const event = ...

emitter.send(event, {
  headers: {
    customheader: "...",
    authorization: "Bearer ...."
  }
})
```

When debugging, it seemed like the passed headers are overwritten by the default headers.

https://github.com/cloudevents/sdk-javascript/blob/99bb88be00f832c58b5fdd467a8a2473830a6547/src/lib/bindings/http/emitter_binary.js#L56
translated to 
> const config = Object.assign(Object.assign({}, options), defaults);

lance:
@c-pius it looks like the spread operator assignment is a bug. Thanks for reporting this.

lholmquist:
I wonder if there are other places that are like this?

lholmquist:
To answer my own question.  It looks like that is the only spot that the ordering was incorrect

Issue #203, opened 2020-06-03
Rename Remaining .js Files to .ts


grant:
It doesn't make sense to have some `.ts` files and some `.js` files in the repo. Although it will work, it is common to just rename the extension of all files to `.ts`.

grant:
Hey @lance, do you have any thoughts around this?

lance:
Hi @grant - I have actually been working on a complete rewrite of the entire module that does a lot more than just renaming the files to have a `.ts` extension. And it's 100% TS. I hope to have at least a draft PR of this ready this week.

grant:
OK, sounds good. Thanks for the update @lance.
The rewrite sounds great and could solve lots of issues.

lance:
I think we can probably consider renaming to 'cloudevents' as well with a 3.0.0 release

grant:
https://www.npmjs.com/~lanceball now has full permissions for module `cloudevents`. Feel free to mirror previous versions/use that module for publishing.

lholmquist:
> https://www.npmjs.com/~lanceball now has full permissions for module cloudevents.
Feel free to mirror previous versions/use that module for publishing.


was it actually decided that this module is changing names?  I thought that was an ongoing discussion in #215 ?

lholmquist:
so getting back to the issue of #225,  if things are being re-written,  should we even worry about the test and just fix the issue now, so the users who reported the bug can do what they need to do and the typescript stuff will fix itself once the module is re-written?

PR #225, opened 2020-06-15
fix: correct the spread order for http_emitter options

lholmquist:
fixes: #223

I would like to add a test for this,  but I think some type definitions need to be added so typescript doesn't complain about me trying to pass in more options in the send method.  Keeping as Draft for now

### DIFF
--- a/src/lib/bindings/http/emitter_binary.js
+++ b/src/lib/bindings/http/emitter_binary.js
@@ -53,7 +53,7 @@ class BinaryHTTPEmitter {
    * @returns {Promise} Promise with an eventual response from the receiver
    */
   async emit(options, cloudevent) {
-    const config = { ...options, ...defaults };
+    const config = { ...defaults, ...options };
     const headers = config[HEADERS];
 
     this.headerParserMap.forEach((parser, getterName) => {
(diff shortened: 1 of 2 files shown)


lholmquist:
While this should be a fairly straightforward change,  i'm running into issues related to Typescript when adding in the tests.

I know basic's, but I'm not sure how to proceed when trying to declare the type for a variable object that uses the spread operator.

@grant perhaps you have some insight, since from my reading of other issues,  you were the most vocal for adding typescript support.   Thanks in advance

Also cc @lance

grant:
That's a bad bug. Thanks for the PR!

I too have seen many TypeScript issues due to the fact this repo is half js, half ts, which causes issues with types due to inferred types.

I've added an issue here: https://github.com/cloudevents/sdk-javascript/issues/203
I'll talk about that issue there.

lance:
I don't know. It could be a while before the rewrite lands and it might be that @c-pius may want to ref `master` for the time being? Care to chime in @c-pius?

c-pius:
From my side there is no urgency. Anyway, thanks for your great work with CE and this SDK!

lholmquist:
closing this since it will be fixed in #226

Issue #219, opened 2020-06-10
CloudEvent Object Shouldn't Create Spec Property


grant:
Currently every new CloudEvent object creates a Specification object.

This doesn't really make sense. A CloudEvent should adhere to a specification, not contain a property called `spec`.

Issue #195, opened 2020-05-29
Formatter should be static


grant:
I noticed the formatter lives within the CloudEvent object.
```js
CloudEvent {
  spec:
   Spec1 {
     payload:
      { specversion: '1.0',
        id: '1234',
        time: '2020-05-28T20:46:18.970Z',
        source: '/some/source',
        type: 'example',
        datacontenttype: undefined,
        data: undefined,
        subject: undefined } },
  formatter: JSONFormatter {},
  extensions: [] }
```
This doesn't really make sense since we can have 1000 CloudEvent objects with the same formatter. 

Maybe it should be a static class/set of methods.

lance:
I was thinking actually of removing it entirely from the class, along with the `format()` function. Instead, expose a top level `Formatter` class or `formatter()` function that takes a `CloudEvent` and formats it as JSON by default. But of course that would be a breaking change and require a version bump. Making it static is a good first step.

lholmquist:
Probably going to hold off on this since it will be part of #226 

should we close?

PR #226, opened 2020-06-16
lib: rewrite in TypeScript

lance:
This is a major rewrite of the entire codebase into TypeScript. Where the code was overly complex, I have attempted to simplify things. A few things remain before it can land.

- [x] Modify linting rules to work better with TS project
- [x] Document the user facing API with ts-doc comments
- [x] Update the README.md with new examples
- [x] Update the example projects in `/examples` to work with the new API
- [x] Convert remainder of JavaScript tests to TypeScript 
- [x] Normalize the use of constants
- [x] Add a test for custom headers (see https://github.com/cloudevents/sdk-javascript/issues/223)
- [x] Add a test for binary data, encoded to base64
- [x] Review `CloudEvent#extensions` and how they are mapped to https://github.com/cloudevents/spec/blob/v1.0/json-format.md 
- [x] Lowercase all CloudEvent property names per spec
- [x] Make HTTP transport mode an enum

Some notes about the changes...

* The generated JS now is written to `./dist`
* The `lib` folder has been removed and everything under it was pulled up into `src`
* The `lib/bindings` folder has been moved to `src/transport`

Fixes: https://github.com/cloudevents/sdk-javascript/issues/223
Fixes: https://github.com/cloudevents/sdk-javascript/issues/219
Fixes: https://github.com/cloudevents/sdk-javascript/issues/203
Fixes: https://github.com/cloudevents/sdk-javascript/issues/195


lholmquist:
I noticed a `_bundles` directory getting created that is not .gitignored,  should it be?

src/event/cloudevent.ts
```diff
@@ -0,0 +1,85 @@
...
+ */
+export class CloudEvent implements CloudEventV1, CloudEventV03 {
+  id: string;
+  type: string;
+  source: string;
+  specversion: Version;
+  dataContentType?: string;
+  dataSchema?: string; // TODO: Is a URI type needed?
+  subject?: string;
+  time?: string|Date;
```
grant, line 22: If we're converting to strings below, we should just use `string`.
If we're allowing `Date`, we should not convert to ISO string below.

```ts
    } else if (this.time instanceof Date) {
      this.time = this.time.toISOString();
    }
```

src/event/cloudevent.ts
```diff
@@ -0,0 +1,85 @@
...
+      throw new TypeError("cannot set dataSchema on version 0.3 event");
+    }
+  }
+
+  /**
+   * Validates this CloudEvent against the schema
+   * @returns boolean
+   * @throws if the CloudEvent does not conform to the schema
+   */
+  validate(): boolean {
```
grant, line 69: Probably should be public.

src/event/cloudevent.ts
```diff
@@ -0,0 +1,85 @@
...
+    return false;
+  }
+
+  public static create(attributes: CloudEventV1Attributes | CloudEventV03Attributes, version?: Version): CloudEvent {
+    const defaults = {
+      specversion: version || Version.V1,
+      id: uuidv4()
+    }
+    return new CloudEvent({ ...defaults, ...attributes });
+  }
```
grant, line 84: Can we just have an overloaded `constructor` (with CloudEvents or CloudEventAttributes) rather than creating a new method?

Maybe for a future PR.

src/event/v03/spec.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+const SUPPORTED_CONTENT_ENCODING = {
+  base64: {
+    check: (data: any) => isBase64(data)
+  }
+};
+
+import { schema } from "./schema";
+import Extensions from "../extensions";
```
grant, line 27: Keep all imports on top of file?

src/event/v03/spec.ts
```diff
@@ -0,0 +1,58 @@
...
+function checkExtensions(extensions: Extensions) {
+  for (const key in extensions) {
+    if (!Object.prototype.hasOwnProperty.call(RESERVED_ATTRIBUTES, key)) {
+      if (isValidType(extensions[key])) return;
+      else throw new ValidationError("Invalid type of extension value");
+    } else {
+      throw new ValidationError(`Reserved attribute name: '${key}'`);
+    }
+  }
+}
```
grant, line 58: We could probably prevent duplicating this code by accepting both `CloudEventV03` and `CloudEventV1`.

src/transport/emitter.ts
```diff
@@ -0,0 +1,64 @@
+import { CloudEvent } from "../event";
+import { emitBinary, emitStructured } from "./http";
+import { Protocol } from ".";
+
+export interface Options {
```
grant, line 5: Could use a more descriptive name.

`TransportOptions`?

src/transport/emitter.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * A class to send binary and structured CloudEvents to a remote endpoint.
+ * Currently, supported protocols are HTTPBinary and HTTPStructured.
+ *
+ * @see https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md
+ * @see https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#13-content-modes
+ */
+export class Emitter {
```
grant, line 28: nit: a name like `CloudEventEmitter` could be more descriptive and help with search.

src/transport/http/binary_emitter.ts
```diff
@@ -0,0 +1,25 @@
...
+import { CloudEvent, Version } from "../../event";
+import { Options } from "../emitter";
+import { Headers, headersFor } from "./headers";
+import { asData } from "../../event/validation";
+
+export async function emitBinary(event: CloudEvent, options: Options): Promise<object> {
+  if (event.specversion !== Version.V1 && event.specversion !== Version.V03) {
+    return Promise.reject(`Unknown spec version ${event.specversion}`)
+  } else {
+    return emit(event, options, headersFor(event));
```
grant, line 12: Seems like this `emit` method could be inlined. We only have a simple `if`/`else` here.

src/transport/http/binary_emitter.ts
```diff
@@ -0,0 +1,25 @@
+import axios, { AxiosRequestConfig } from "axios";
+
+import { CloudEvent, Version } from "../../event";
+import { Options } from "../emitter";
+import { Headers, headersFor } from "./headers";
+import { asData } from "../../event/validation";
+
+export async function emitBinary(event: CloudEvent, options: Options): Promise<object> {
```
grant, line 8: Comment?
Makes a POST request with the CloudEvent to...

src/transport/http/headers.ts
```diff
@@ -0,0 +1,90 @@
...
+import { MappedParser } from "../../parsers";
+
+export interface Headers {
+  [key: string]: string
+};
+
+export const allowedContentTypes = ["application/json", "application/octet-stream"];
+export const requiredHeaders = [ "ce-id", "ce-source", "ce-type", "ce-specversion" ];
+
+export function validate(headers: Headers): Headers {
```
grant, line 14: Comment?

src/transport/http/v03/headers.ts
```diff
@@ -0,0 +1,35 @@
...
+export const headerMap: Map<string, MappedParser> = new Map();
+headerMap.set("dataContentType", parser(HEADER_CONTENT_TYPE));
+headerMap.set("dataContentEncoding", parser(CONTENT_ENCODING));
+headerMap.set("subject", parser(SUBJECT));
+headerMap.set("type", parser(TYPE));
+headerMap.set("specversion", parser(SPEC_VERSION));
+headerMap.set("source", parser(SOURCE));
+headerMap.set("id", parser(ID));
+headerMap.set("time", parser(TIME));
+headerMap.set("schemaURL", parser(SCHEMA_URL));
```
grant, line 35: Let's use a raw JavaScript object instead of a Map. If there is a very strong reason for using a Map, follow the below suggestion.

---

Let's use a `Map` constructor with key value pairs rather than using the `set` method N times.

Example:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/Map

src/transport/http/v03/parsers.ts
```diff
@@ -0,0 +1,35 @@
...
+const structuredParsers: Map<string, MappedParser> = new Map();
+structuredParsers.set("type", parser("type"));
+structuredParsers.set("specversion", parser("specversion"));
+structuredParsers.set("source", parser("source"));
+structuredParsers.set("id", parser("id"));
+structuredParsers.set("time", parser("time", new DateParser()));
+structuredParsers.set("schemaurl", parser("schemaURL"));
+structuredParsers.set("datacontentencoding", parser("dataContentEncoding"));
+structuredParsers.set("subject", parser("subject"));
+structuredParsers.set("data", parser("data"));
```
grant, line 30: Same suggestion for using a regular JavaScript object. It should be more convenient and performant.

src/transport/http/headers.ts
```diff
@@ -0,0 +1,90 @@
...
+import { headerMap as v03Map } from "./v03";
+import { Version } from "../../event";
+import { MappedParser } from "../../parsers";
+
+export interface Headers {
+  [key: string]: string
+};
+
+export const allowedContentTypes = ["application/json", "application/octet-stream"];
+export const requiredHeaders = [ "ce-id", "ce-source", "ce-type", "ce-specversion" ];
```
grant, line 12: We use the strings "ce-id", etc. many times.

Do we want to create a `v1/constants.ts` file or something equivalent and export constants with string from there?

src/transport/protocols.ts
```diff
@@ -0,0 +1 @@
+export const enum Protocol { HTTPBinary, HTTPStructured, HTTP };
```
grant, line 1: For small files like these, it might be worth it to inline it rather than have a new file.

For example, we can inline the contents into the `index.ts` file.

src/transport/receiver.ts
```diff
@@ -45,44 +49,42 @@ export class HTTPReceiver {
...
   throw new ValidationError("no cloud event detected");
 }
 
-function getVersion(mode: string, headers: { [key: string]: string }, body: string | { specversion?: string }) {
-  if (mode === BINARY) {
+function getVersion(mode: string, headers: Headers, body: string | object | CloudEventV03 | CloudEventV1) {
+  if (mode === "binary") {
     // Check the headers for the version
-    const versionHeader = headers[DEFAULT_SPEC_VERSION_HEADER];
+    const versionHeader = headers["ce-specversion"];
```
grant, line 79: I think keeping the constants (although probably renamed) would be useful rather than raw strings.

src/transport/receiver.ts
```diff
@@ -45,44 +49,42 @@ export class HTTPReceiver {
...
+  if (contentType && contentType.startsWith("application/cloudevents")) {
+    return "structured";
   }
+  if (headers["ce-id"]) { return "binary"; }
   throw new ValidationError("no cloud event detected");
 }
 
-function getVersion(mode: string, headers: { [key: string]: string }, body: string | { specversion?: string }) {
-  if (mode === BINARY) {
+function getVersion(mode: string, headers: Headers, body: string | object | CloudEventV03 | CloudEventV1) {
```
grant, line 76: Mode should probably be an enum.

test-ts/http_receiver_test.ts
```diff
@@ -83,24 +69,20 @@ describe("HTTP Transport Binding Receiver for CloudEvents", () => {
...
-        [DEFAULT_SPEC_VERSION_HEADER]: specversion,
-        [BINARY_HEADERS_03.ID]: id,
-        [BINARY_HEADERS_03.TYPE]: type,
-        [BINARY_HEADERS_03.SOURCE]: source
+      const binaryHeaders = {
+        "content-type": "application/json; charset=utf-8",
+        "ce-specversion": specversion,
+        "ce-id": id,
+        "ce-type": type,
+        "ce-source": source
```
grant, line 82: As mentioned, I like the other style before as it prevents typos and allows for autocompletion.

package.json
```diff
@@ -97,6 +95,7 @@
   },
   "homepage": "https://github.com/cloudevents/sdk-javascript#readme",
   "dependencies": {
+    "@types/uuid": "^8.0.0",
```
grant, line 98: `@types/*` should be a devDep.

lance:
@lholmquist good catch! That's actually a bug on master. :bug:

src/transport/receiver.ts
lance: Agree - I just hadn't normalized my usage of the `constants.ts` file yet. That's in the TODO list in the PR description. I didn't like the way I was having to import them, so I punted and used strings until I had time to go back and figure out the best way to surface all of them. I personally don't think there should be a single `constants.ts` file for all constants in the project. Rather, I think they should be grouped accordingly - e.g. all of the header constants in `transport/http` somewhere.

src/event/cloudevent.ts
lance: The `time` property, per the [spec MUST adhere to RFC3339](https://github.com/cloudevents/spec/blob/v1.0/spec.md#time). This enforces that.

src/event/cloudevent.ts
lance: It was that way and now I don't remember why I felt compelled to change it. I'll look at it again.

src/transport/http/binary_emitter.ts
```diff
@@ -0,0 +1,25 @@
...
+    return emit(event, options, headersFor(event));
+  }
+}
+
+async function emit(event: CloudEvent, options: Options, headers: Headers): Promise<object> {
+  const contentType: Headers = { "content-type": "application/json; charset=utf-8" };
+  const config = {
+    ...options,
+    method: "POST",
+    headers: { ...contentType, ...headers },
```
lholmquist, line 21: I sent this PR to your branch, https://github.com/lance/sdk-javascript/pull/1 

If a custom header is passed in the options like 
```
{ headers: { customHeader: "value" } }
```
It was actually getting overwritten since the spread operator does a shallow merge(i think)

lance: Thanks! I've merged that change. :)

src/event/cloudevent.ts
```diff
@@ -0,0 +1,122 @@
...
+  data_base64?: any;
+  [key: string]: any; // Extensions should not exist as it's own object, but instead be keyed directly from the event
+
+  // V03 deprecated attributes
+  schemaurl?: string;
+  datacontentencoding?: string;
+
+  constructor(event: CloudEventV1 | CloudEventV1Attributes | CloudEventV03 | CloudEventV03Attributes) {
+    // copy the incoming event so that we can delete properties as we go
+    // everything left after we have deleted know properties becomes an extension
```
lholmquist, line 34: what is the reasoning for not keeping extensions its own object within the event?

lance: Because [in the spec](https://github.com/cloudevents/spec/blob/master/primer.md#json-extensions), when a CloudEvent is serialized to for wire transport, the extensions are on the main event itself and not in an `extensions` object.

> As mentioned in the Attributes section of the JSON Event Format for CloudEvents specification, CloudEvent extension attributes are serialized as siblings to the specification defined attributes - meaning, at the top-level of the JSON object. 

With the previous implementation, there was a `CloudEvent.format()` function that would serialize things properly. But one of the goals of this big rewrite was to make it so that a `CloudEvent` could be serialized for transport with little more than `JSON.stringify(event)`, and that a new event could be created with the extensions inline, e.g. `new CloudEvent({ extension: 'some value'});`.

lholmquist: sounds good.  I do remember seeing an article why using `delete` might not be as efficient, but of course i can't find it,  and it was probably more than 5 yrs old,  so it is possible that info could be outdated

package.json
```diff
@@ -2,26 +2,22 @@
...
   "scripts": {
     "watch": "tsc --project tsconfig.json --watch",
     "build": "tsc --project tsconfig.json && tsc --project tsconfig.browser.json && webpack",
-    "prelint": "npm run build",
-    "lint": "standardx src examples",
-    "fix": "standardx --fix",
-    "pretest": "npm run lint && npm run test:ts",
-    "test": "mocha test/**/*.js",
-    "test:ts": "mocha --require ts-node/register ./test-ts/**/*.ts",
+    "lint": "eslint 'src/**/*.{js,ts}' 'test/**/*.{js,ts}' --fix",
```
lholmquist, line 9: Using the `--fix` flag might be ok,  but if someone submits a PR and doesn't do the `npm run test` locally first we would never know if there was linting issues since it would just get fixed on CI,  but not committed back

lance: Ahh good point.

grant: I usually like `npm run fix` along with `npm run lint`.

lance:
@cloudevents/sdk-javascript-maintainers I have completed all of the items on the checklist and addressed about 95% of the comments/nits. Please have a look. This is ready for final review and landing. :airplane:

src/transport/http/v03/parsers.ts
```diff
@@ -0,0 +1,35 @@
...
+structuredParsers.set(CONSTANTS.CE_ATTRIBUTES.ID, parser(CONSTANTS.CE_ATTRIBUTES.ID));
+structuredParsers.set(CONSTANTS.CE_ATTRIBUTES.TIME, parser(CONSTANTS.CE_ATTRIBUTES.TIME, new DateParser()));
+structuredParsers.set(CONSTANTS.STRUCTURED_ATTRS_03.SCHEMA_URL, parser(CONSTANTS.STRUCTURED_ATTRS_03.SCHEMA_URL));
+structuredParsers.set(
+  CONSTANTS.STRUCTURED_ATTRS_03.CONTENT_ENCODING,
+  parser(CONSTANTS.STRUCTURED_ATTRS_03.CONTENT_ENCODING),
+);
+structuredParsers.set(CONSTANTS.CE_ATTRIBUTES.CONTENT_TYPE, parser(CONSTANTS.CE_ATTRIBUTES.CONTENT_TYPE));
+structuredParsers.set(CONSTANTS.CE_ATTRIBUTES.SUBJECT, parser(CONSTANTS.CE_ATTRIBUTES.SUBJECT));
+structuredParsers.set(CONSTANTS.CE_ATTRIBUTES.DATA, parser(CONSTANTS.CE_ATTRIBUTES.DATA));
```
grant, line 33: I'm not sure what all of this is doing, but it seems over-the-top.

Why can't we use a normal map (`{}`) rather than `Map`?

src/transport/receiver.ts
```diff
@@ -0,0 +1,104 @@
...
+      case Version.V03:
+        return this.receivers.v03[mode].parse(body, headers);
+      default:
+        console.error(`Unknown spec version ${version}. Default to ${Version.V1}`);
+        return this.receivers.v1[mode].parse(body, headers);
+    }
+  }
+}
+
+function getMode(headers: Headers): Mode {
```
grant, line 77: Add a comment?

grant:
Looks good! 👍 A few comments.

src/event/cloudevent.ts
```diff
@@ -0,0 +1,168 @@
...
+    const properties = { ...event };
+
+    this.id = (properties.id as string) || uuidv4();
+    delete properties.id;
```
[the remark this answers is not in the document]
lance: As [noted in the comment above](https://github.com/nodeshift/opossum/blob/master/LICENSE), it is necessary to delete them if we want to have extensions as siblings of the event properties. We remove all of the known properties, and [what's left are the extensions](https://github.com/cloudevents/sdk-javascript/pull/226/files#diff-112b25e99d71e1fed6e04eb3f3e6535eR97-R100).

src/transport/receiver.ts
lance: Sure - but you realize this isn't part of the public facing API, right?

src/transport/http/binary_emitter.ts
```diff
@@ -0,0 +1,31 @@
...
+
+async function emit(event: CloudEvent, options: TransportOptions, headers: Headers): Promise<AxiosResponse> {
+  const contentType: Headers = { [CONSTANTS.HEADER_CONTENT_TYPE]: CONSTANTS.DEFAULT_CE_CONTENT_TYPE };
+  const config = {
+    ...options,
+    method: "POST",
+    headers: { ...contentType, ...headers, ...(options.headers as Headers) },
+    data: asData(event.data, event.datacontenttype as string),
+  };
+  return axios.request(config as AxiosRequestConfig);
```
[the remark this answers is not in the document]
lance: I find it more readable this way

README.md
```diff
@@ -36,73 +36,70 @@ binary and structured events in either the 1.0 or 0.3 protocol formats.
...
   .then((response) => {
     // handle the response
   }).catch(console.error);
-
 ```
 
+### Example Applications
+
+There are a few trivial example applications in
+[the examples folder](https://github.com/cloudevents/sdk-javascript/tree/master/examples).
```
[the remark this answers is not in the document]
lance: That doesn't work so well with the readme also published to npmjs.com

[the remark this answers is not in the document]
lance: That doesn't work so well when the readme is also published on npmjs.com

src/event/v03/spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (!isValidAgainstSchema(event)) {
+    throw new ValidationError("invalid payload", isValidAgainstSchema.errors);
+  }
+  return checkDataContentEncoding(event);
+}
+
+function checkDataContentEncoding(event: CloudEventV03): boolean {
+  if (event.datacontentencoding) {
+    // we only support base64
+    const encoding = event.datacontentencoding.toLocaleLowerCase("en-US");
```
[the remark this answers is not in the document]
lance: Good catch. Thanks!

src/transport/http/structured_receiver.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    // Clone and low case all headers names
+    const sanitizedHeaders = sanitize(headers);
+
+    const contentType = sanitizedHeaders[CONSTANTS.HEADER_CONTENT_TYPE];
+    const parser: Parser = contentType ? parserByContentType[contentType] : new JSONParser();
+    if (!parser) throw new ValidationError(`invalid content type ${sanitizedHeaders[CONSTANTS.HEADER_CONTENT_TYPE]}`);
+    const incoming = { ...(parser.parse(payload) as Record<string, unknown>) };
+    // eslint-disable-next-line no-console
+    console.log(incoming);
```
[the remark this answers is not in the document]
lance: I hate it when I do that :)

lance:
Deep breaths.... landing this.

PR #227, opened 2020-06-17
chore: add genertated _bundles to gitignore

lholmquist:
the `_bundles/` was not ignored


lance:
@lholmquist the reason this didn't get ignored is because I wanted to change it to `/bundles` and did so in `.gitignore`, but forgot to change it in `webpack.config.js`. Would you mind modifying this PR to actually change the output directory to `bundles`?

lholmquist:
ah,  sure

lholmquist:
updated

Issue #230, opened 2020-06-22
Receiver accept() method modifies its input parameter.


matejvasek:
The receiver may destroy input payload.
For instance after calling `receiver.accept(response.headers, response.data)` the data member may end up as empty object. 

https://github.com/cloudevents/sdk-javascript/blob/fba3294ce04a30be0e5ab551a1fa01727dc8d1f8/src/lib/bindings/http/validation/structured.js

lholmquist:
I'm not totally following from the link you posted.   Would it be possible to post a small code example of what you are seeing?

lance:
@lholmquist the issue is here: https://github.com/cloudevents/sdk-javascript/blob/fba3294ce04a30be0e5ab551a1fa01727dc8d1f8/src/lib/bindings/http/validation/structured.js#L38. Even though the `payload` is assumed to be transformed by the parser [here][(https://github.com/cloudevents/sdk-javascript/blob/fba3294ce04a30be0e5ab551a1fa01727dc8d1f8/src/lib/bindings/http/validation/structured.js#L29) the parser in many cases is a "pass through" parser - doing nothing but returning the parameter. So we end up with the caller's `payload` object being modified.

lholmquist:
ah,  ok.  sounds like a bug

matejvasek:
@lance @lholmquist  note that I reproduced this on the Lance's TS rework branch, I am going to try to reproduce this on master.

matejvasek:
Yes, the "pass through" parser causes problem.

matejvasek:
I would submit PR, but I won't since @lance is going to rewrite a lot of things.

PR #231, opened 2020-06-24
fix: parse method mutating its input

matejvasek:
fixes: #230

### DIFF
--- a/src/lib/bindings/http/validation/structured.js
+++ b/src/lib/bindings/http/validation/structured.js
@@ -26,7 +26,7 @@ function parse(payload, headers, receiver) {
   const sanityHeaders = sanityAndClone(headers);
   const contentType = sanityHeaders[HEADER_CONTENT_TYPE];
   const parser = receiver.parserByMime[contentType];
-  const incoming = parser.parse(payload);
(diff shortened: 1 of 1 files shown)


src/lib/bindings/http/validation/structured.js
```diff
@@ -32,17 +32,20 @@ function parse(payload, headers, receiver) {
     source: undefined
   };
 
+
+  const unprocessed = new Set(Object.keys(incoming));
```
lholmquist, line 36: is `Set` necessary here.   Can't we assume that the keys in the object are unique already?

matejvasek: I think they should be unique. The thing is: If I understand it correctly `Object.keys(incoming)` returns an array and deletion from an array is more complicated than just `delete()` which is offered by a `Set` object.

lance, line 36: I think the most efficient would be using the spread operator like this.
```suggestion
  const unprocessed = { ...incoming };
```

matejvasek:
@lance I force pushed it. There is now only one line changed: `const incoming = { ...parser.parse(payload) };`.

Issue #29, opened 2019-08-07
Immutable event builder


fabiojose:
Implements an immutable event builder, to builds events instances that are unchangeable after its build.

Depends on #28

lance:
@fabiojose I agree that events should probably be immutable - aside from a well-defined path for converting from/to spec versions. Given how the repo has been evolving, how do you think this issue should progress?

grant:
> how do you think this issue should progress?

If we use JSON for CloudEvents, then we can just `Object.freeze` our event data to make it immutable.

Requires https://github.com/cloudevents/sdk-javascript/issues/65.

lance:
I'm going to mark this a stalled for now. TypeScript users now have `Object.freeze` for regular objects. However, that doesn't work for the `CloudEvent` class. We'd need to add a `freeze()` method to the class itself to prevent modification. This is because the setters on the class [delegate to `this.spec`](https://github.com/cloudevents/sdk-javascript/blob/master/src/lib/cloudevent.ts#L109-L111).

grant:
Marking all `CloudEvent` fields as `readonly` should be a solution.

This is very common and recommended in TypeScript. See vscode's source for examples.

lance:
Weird that the bot closed this

lance:
While not exposed via an actual _builder_, I think https://github.com/cloudevents/sdk-javascript/pull/234 should satisfy this issue. Closing.

Issue #229, opened 2020-06-20
CloudEvent extensions should be restricted to types allowed in spec


lance:
Currently an extension can have any value - e.g. it could be an object. However, the spec limits what value types an extension can have. See https://github.com/cloudevents/spec/blob/master/spec.md#type-system

lholmquist:
Since the spec doesn't allow an Object, should we be throwing an error if someone passes in this:
```
new CloudEvent({
...
...
extension1: { key: value }
})
```
Or is this ok since it will get converted to a JSON before transmission.  basically,  should this test pass or fail:  https://github.com/cloudevents/sdk-javascript/blob/master/test/spec_1_tests.ts#L99

lance:
That test should probably fail. We can't assume that the receiver will know that the value needs to be converted to an object. Since strings are valid value types, the user should use `JSON.stringify` before setting the extension.

Issue #228, opened 2020-06-20
CloudEvent should only allow [^a-z|0-9$] for extension names


lance:
Per the spec's [naming conventions](https://github.com/cloudevents/spec/blob/master/spec.md#attribute-naming-convention), attribute names should be all lower case, with nothing but a-z and 0-9 in the name. This module currently does not enforce that.

lholmquist:
Is the "-" also allowed.  reading the spec,  it doesn't look like it.  If that is the case, then we will need to update the examples

lance:
It is not allowed. Only alphanumeric characters.

> CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be descriptive and terse and SHOULD NOT exceed 20 characters in length.

https://github.com/cloudevents/spec/blob/master/spec.md#attribute-naming-convention

lholmquist:
Should we make this a semver-major change then?

lance:
> Should we make this a semver-major change then?

Yes - we will have a semver major release as soon as the typescript version lands, so it should land with that.

PR #234, opened 2020-06-30
Validate a Cloud Event on creation and make it read only with a method to augment it

### LINKED ISSUE #233: A CloudEvent should be readonly but provide a way to augment itself
With recent API changes, a `CloudEvent` is nicely represented as a plain old JS object. For example, for most purposes, this is actually a valid CE.

```js
const ce = {
  source: '/',
  type: 'example',
  specversion: 'v1.0',
  id: 'asd0aan3412juv'
}
```

However, because we want to have spec compliance checks, and provide default values (e.g. a UUID or timestamp), we have a constructor and a class. The example above can be written as:

```js
const ce = new CloudEvent({
  source: '/',
  type: 'example',
});
```

Here, code in the constructor is providing default values for `id` and `specversion`.

However this introduces some tricky edge cases. For example, extensions. According to the spec, extensions must exist as siblings of the CE properties. And JS being as loosey goosey as it is, this means a user can do some things outside of the spec boundaries. For example

```js
const ce = new CloudEvent({ source: '/', type: 'example' });
ce[extension-1] = // <- note invalid extension name
  {
    foo: 'bar' // <-- note invalid extension value
  };
```

A solution to this is to

1. At the bottom of the `CloudEvent` constructor, call `this.validate()`
2. Make the `CloudEvent` object read only just after validation with `Object.freeze(this);`
3. Provide a "builder" of some kind that enables cloning an event with new properties/extensions, e.g. `myEvent.cloneWith( { extension: 'value' });` (this could be a function on `CloudEvent` or a class unto itself.

Note that both #228 and #229 should be implemented and added to the validation step as a part of this.

Related: https://github.com/cloudevents/sdk-javascript/issues/29

lholmquist:
- [x] Validate the Cloud Event in the Constructor
- [x] Freeze the Object that is created
- [x] Update/Add extensions after the fact
- [x] Modify other properties if needed
- [x] Update Tests
 

During the creation of a CloudEvent,  the validate method is now called.  The object is also frozen to make it read-only.

A new method `cloneWith` has been added which can be used to add/update attributes such as extensions and will return a new Cloud Event Object

fixes #233 

related #29 

### DIFF
--- a/src/event/cloudevent.ts
+++ b/src/event/cloudevent.ts
@@ -1,7 +1,7 @@
 import { v4 as uuidv4 } from "uuid";
 
-import { CloudEventV1, validateV1, CloudEventV1Attributes } from "./v1";
-import { CloudEventV03, validateV03, CloudEventV03Attributes } from "./v03";
+import { CloudEventV1, validateV1, CloudEventV1Attributes, CloudEventV1OptionalAttributes } from "./v1";
+import { CloudEventV03, validateV03, CloudEventV03Attributes, CloudEventV03OptionalAttributes } from "./v03";
 import { ValidationError, isBinary, asBase64 } from "./validation";
 import CONSTANTS from "../constants";
 import { isString } from "util";
@@ -98,6 +98,10 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
     for (const [key, value] of Object.entries(properties)) {
       this[key] = value;
     }
+
+    this.validate();
+
+    Object.freeze(this);
   }
 
   get time(): string | Date {
@@ -164,5 +168,23 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
         throw new ValidationError("invalid payload", e);
       }
     }
+  }
+
+  /**
+   * Clone a CloudEvent with new/update attributes
+   * @param {object} options attributes to augment the CloudEvent with
+   * @throws if the CloudEvent does not conform to the schema
+   * @return {CloudEvent} returns a new CloudEvent
+   */
+  public cloneWith(
+    options:
+      | CloudEventV1
+      | CloudEventV1Attributes
+      | CloudEventV1OptionalAttributes
+      | CloudEventV03
+      | CloudEventV03Attributes
+      | CloudEventV03OptionalAttributes,
+  ): CloudEvent {
+    return new CloudEvent(Object.assign({}, this.toJSON(), options) as CloudEvent);
   }
 }
--- a/src/event/v1/cloudevent.ts
+++ b/src/event/v1/cloudevent.ts
@@ -25,7 +25,7 @@ export interface CloudEventV1 extends CloudEventV1Attributes {
   specversion: string;
 }
 
-export interface CloudEventV1Attributes {
+export interface CloudEventV1Attributes extends CloudEventV1OptionalAttributes {
   /**
    * [REQUIRED] Identifies the context in which an event happened. Often this
    * will include information such as the type of the event source, the
@@ -58,7 +58,9 @@ export interface CloudEventV1Attributes {
    * @example com.example.object.delete.v2
    */
   type: string;
+}
 
+export interface CloudEventV1OptionalAttributes {
   /**
    * The following fields are optional.
    */
--- a/src/event/v03/cloudevent.ts
+++ b/src/event/v03/cloudevent.ts
@@ -25,7 +25,7 @@ export interface CloudEventV03 extends CloudEventV03Attributes {
   specversion: string;
 }
 
-export interface CloudEventV03Attributes {
+export interface CloudEventV03Attributes extends CloudEventV03OptionalAttributes {
   /**
    * [REQUIRED] Identifies the context in which an event happened. Often this
    * will include information such as the type of the event source, the
@@ -57,7 +57,9 @@ export interface CloudEventV03Attributes {
    * @example com.example.object.delete.v2
    */
   type: string;
+}
 
+export interface CloudEventV03OptionalAttributes {
   /**
    * The following fields are optional.
    */
--- a/test/integration/cloud_event_test.ts
+++ b/test/integration/cloud_event_test.ts
@@ -139,7 +139,7 @@ describe("A 0.3 CloudEvent", () => {
   });
 
   it("can be constructed with a datacontentencoding", () => {
-    const ce = new CloudEvent({ datacontentencoding: "Base64", ...v03fixture });
+    const ce = new CloudEvent({ datacontentencoding: "Base64", ...v03fixture, data: "SSB3YXMgZnVubnkg8J+Ygg==" });
     expect(ce.datacontentencoding).to.equal("Base64");
   });
 
--- a/test/integration/http_binding_03.ts
+++ b/test/integration/http_binding_03.ts
@@ -38,9 +38,9 @@ const cloudevent = new CloudEvent({
   dataschema: "",
   datacontentencoding: "",
   data_base64: "",
+  [ext1Name]: ext1Value,
+  [ext2Name]: ext2Value,
 });
-cloudevent[ext1Name] = ext1Value;
-cloudevent[ext2Name] = ext2Value;
 
 const cebase64 = new CloudEvent({
   specversion: Version.V03,
@@ -51,9 +51,9 @@ const cebase64 = new CloudEvent({
   time,
   schemaurl,
   data: dataBase64,
+  [ext1Name]: ext1Value,
+  [ext2Name]: ext2Value,
 });
-cebase64[ext1Name] = ext1Value;
-cebase64[ext2Name] = ext2Value;
 
 const webhook = "https://cloudevents.io/webhook";
 const httpcfg = {
--- a/test/integration/http_binding_1.ts
+++ b/test/integration/http_binding_1.ts
@@ -25,7 +25,7 @@ const ext1Value = "foobar";
 const ext2Name = "extension2";
 const ext2Value = "acme";
 
-const cloudevent = new CloudEvent({
+let cloudevent = new CloudEvent({
   specversion: Version.V1,
   type,
   source,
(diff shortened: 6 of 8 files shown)


lholmquist:
I just added Initial support for validating with the Proxy.  Some tests are still failing, but I would like to talk about this Proxy handler:

```
return new Proxy(this, {
      // obj is the instance of the CloudEvent
      // prop is the property being set
      // value is the value on the right side of the equal sign
      set: function (obj, prop: string, value) {
        // Make a copy of the incoming Object, need to have all it's properties and such
        const updateObj = Object.create(obj);
        // Update it with the new value
        updateObj[prop] = value;

        // Validate the object
        obj.validate(updateObj);

        // If we succeed, then Update the real object
        // Set the new value normally
        obj[prop] = value;

        return true;
      },
    });
```

At first the handler was simpler and looked like this :
```
set: function (obj, prop, vale) {
  obj[prop] = value;
  return this.validate()
}
```

The only problem with this, is if a validation fails,  the value passed in was still being assigned.  This was causing some tests to fail because we are re-using the same CloudEvent Object across multiple tests.   In a real application it might not be an issue, since if an invalid value is assigned,  an error is thrown.   So not sure if this is overkill and we should re-write the tests or if we should try to validate before assigning.  Hopefully what i just rambled out makes sense :) 

Also the reason i'm using `Object.create` and not `...` is because `...` strips off  the set and get methods of time and data and we need those.

lholmquist:
Going in a different direction. Not using Proxy any more but using the "cloneWith" method instead.

I'll need to update the tests

src/event/cloudevent.ts
```diff
@@ -165,4 +169,8 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
       }
     }
   }
+
+  public cloneWith(options?: any): CloudEvent {
```
lance, line 173: You should type this parameter like this:

```suggestion
  public cloneWith(event: CloudEventV1 | CloudEventV1Attributes | CloudEventV03 | CloudEventV03Attributes): CloudEvent {
```

lholmquist: yea,  it was a placeholder for the moment.  But i don't think this should be a CloudEvent Type.  The value passed in could just be a couple extensions like this : `{ ext1: "value1", ext2: "value2" }`  and it Typscript will error in that case.

lholmquist: reading what i wrote,  i don't think i explained it well enough.

I think we should be able to do this:

```
// Initial Creation
let cloudevent = new CloudEvent({....});

// Update with some extensions
cloudevent = cloudevent.cloneWith({ ext1: "value1", ext2: "value2"  });
```

You can't do that update if the function parameter is of the CloudEvent type.   I was assuming this was how it should work,  you pass in the things you want to add.

lholmquist: just updated with the suggestions we talked about

src/event/v1/cloudevent.ts
```diff
@@ -58,7 +58,9 @@ export interface CloudEventV1Attributes {
    * @example com.example.object.delete.v2
    */
   type: string;
+}
 
+export interface CloudEventV1OptionsAttributes {
```
lance, line 63: Can you name this `CloudEventV1OptionalAttributes` instead?

lholmquist: right,  that name makes more sense

lholmquist:
For this PR i had originally thought to add the validation in #228 and #229, but i think those should be its own PR and this one should just focus on calling validate during object creation as well as making the object read-only.


I have another branch with the extension name validation ready to go once this gets in

lholmquist:
Ping @cloudevents/sdk-javascript-maintainers PTAL

test/http_binding_1.ts
```diff
@@ -35,8 +35,7 @@ const cloudevent = new CloudEvent({
   dataschema,
   data,
 });
-cloudevent[ext1Name] = ext1Value;
-cloudevent[ext2Name] = ext2Value;
+cloudevent = cloudevent.cloneWith({ [ext1Name]: ext1Value, [ext2Name]: ext2Value });
```
grant, line 38: Why can't we just add this to the original CloudEvent?

test/cloud_event_test.ts
```diff
@@ -139,7 +139,7 @@ describe("A 0.3 CloudEvent", () => {
   });
 
   it("can be constructed with a datacontentencoding", () => {
-    const ce = new CloudEvent({ datacontentencoding: "Base64", ...v03fixture });
+    const ce = new CloudEvent({ datacontentencoding: "Base64", ...v03fixture, data: "SSB3YXMgZnVubnkg8J+Ygg==" });
```
grant, line 142: Can you use a function to Base64 encode the string rather than post the literal?

test/spec_03_tests.ts
```diff
@@ -46,8 +46,13 @@ describe("CloudEvents Spec v0.3", () => {
 
   describe("OPTIONAL Attributes", () => {
     it("Should have 'datacontentencoding'", () => {
-      cloudevent.datacontentencoding = Constants.ENCODING_BASE64;
+      cloudevent = cloudevent.cloneWith({
+        datacontentencoding: Constants.ENCODING_BASE64,
+        data: "SSB3YXMgZnVubnkg8J+Ygg==",
```
grant, line 51: Same here, can you post the original string, then `btoa` it?

test/spec_1_tests.ts
```diff
@@ -14,7 +14,7 @@ const data = {
 };
 const subject = "subject-x0";
 
-const cloudevent = new CloudEvent({
+let cloudevent = new CloudEvent({
```
grant, line 17: Don't use a `let`.
Use a `const` and create new `const`s if needed.

test/spec_1_tests.ts
```diff
@@ -65,30 +65,30 @@ describe("CloudEvents Spec v1.0", () => {
 
   describe("Extensions Constraints", () => {
     it("should be ok when type is 'boolean'", () => {
-      cloudevent["ext-boolean"] = true;
+      cloudevent = cloudevent.cloneWith({ "ext-boolean": true });
       expect(cloudevent.validate()).to.equal(true);
```
grant, line 69: This can probably be inlined:

```ts
expect(
  cloudevent.cloneWith({ "ext-boolean": true }).validate()
).to.be.true;

File: test/http_binding_1.ts (hunk 3 above)
Comment (lholmquist) [reply test/http_binding_1.ts:38]: probably could.  Looks like a leftover of when you couldn't add extensions during construction

File: test/spec_1_tests.ts (hunk 6 above)
Comment (lholmquist) [reply test/spec_1_tests.ts:17]: Doing it this way was the path of least resistance to updating the tests now that the CloudEvent is frozen.  Doing it this way also allows us to test the cloneWith method without having to write a bunch more tests.

File: test/spec_1_tests.ts (hunk 7 above)
Comment (lholmquist) [reply test/spec_1_tests.ts]: yup

File: test/spec_03_tests.ts (hunk 5 above)
Comment (lholmquist) [reply test/spec_03_tests.ts:51]: There are other tests that already have the string encoded,  so the precedent was already set.  If this is a blocker, then we can create another PR to update the other tests as well

**Comment by lholmquist** [comment]
@cloudevents/sdk-javascript-maintainers how do we feel about merging this?

File: test/spec_1_tests.ts [hunk 8]
```diff
@@ -97,73 +97,60 @@ describe("CloudEvents Spec v1.0", () => {
...
 
     describe("'source'", () => {
-      it("should throw an error when is absent", () => {
-        delete cloudevent.source;
-        expect(cloudevent.validate.bind(cloudevent)).to.throw(ValidationError, "invalid payload");
-        cloudevent.source = source;
+      it("should throw an error when trying to remove", () => {
+        expect(() => {
+          delete cloudevent.source;
+        }).to.throw();
```
lance, line 117:
```suggestion
        }).to.throw(ValidationError, "invalid payload");
```

test/spec_1_tests.ts
```diff
@@ -97,73 +97,60 @@ describe("CloudEvents Spec v1.0", () => {
...
-      });
-
-      it("should throw an error when is empty", () => {
-        cloudevent.specversion = "" as Version;
-        expect(cloudevent.validate.bind(cloudevent)).to.throw(ValidationError, "invalid payload");
-        cloudevent.specversion = Version.V1;
+      it("should throw an error when trying to remove", () => {
+        expect(() => {
+          delete cloudevent.specversion;
+        }).to.throw();
```
lance, line 125:
```suggestion
        }).to.throw(ValidationError, "invalid payload");
```

test/spec_1_tests.ts
```diff
@@ -97,73 +97,60 @@ describe("CloudEvents Spec v1.0", () => {
...
-      });
-
-      it("should throw an error when is an empty string", () => {
-        cloudevent.type = "";
-        expect(cloudevent.validate.bind(cloudevent)).to.throw(ValidationError, "invalid payload");
-        cloudevent.type = type;
+      it("should throw an error when trying to remove", () => {
+        expect(() => {
+          delete cloudevent.type;
+        }).to.throw();
```
lance, line 133:
```suggestion
        }).to.throw(ValidationError, "invalid payload");
```

test/spec_1_tests.ts
lance: Agree - this is exercising the `cloneWith` function, so I think is good as-is.

test/spec_1_tests.ts
lholmquist: They actually throw a TypeError instead of a ValidationError,  since the error now comes from deleting a property on a frozen object.

I can still update the tests to check for the TypeError Throw

lholmquist:
@lance i updated the tests to make sure they are throwing a TypeError, which is the error that will be thrown when trying to delete a property on a frozen object.

lholmquist:
@grant are you ok with this PR.  Your comments were related to how some of the tests were written, which i've tried to give an explanation for.  If this is still a blocker, maybe we could address those in a separate PR?

test/spec_03_tests.ts
grant: Maybe add a TODO if other PRs are doing this?
It's still not ideal as I can't read or understand SSB3...

test/spec_1_tests.ts
grant: Sounds fine.

grant:
Looks good to me. Thanks for the PR!

PR #236, opened 2020-07-02
feat: implementation of discovery spec

loopingz:
See: https://github.com/cloudevents/spec/blob/master/discovery.md

This PR still need the unit testing but it is pushed for discussion in our weekly call

### DIFF
--- a/examples/express-ex/index.js
+++ b/examples/express-ex/index.js
@@ -1,10 +1,49 @@
 /* eslint-disable */
 
 const express = require("express");
-const { Receiver } = require("cloudevents");
+const { Receiver, DiscoveryService } = require("cloudevents");
 const app = express();
-const bodyParser = require('body-parser')
-app.use(bodyParser.json())
+const bodyParser = require("body-parser");
+app.use(bodyParser.json());
+
+DiscoveryService.registerService({
+  url: "https://example.com/services/widgetService",
+  name: "com.example.services.widgetService",
+  specversions: ["1.0"],
+  subscriptionurl: "https://events.example.com",
+  protocols: ["HTTP"],
+  types: [
+    {
+      type: "com.example.widget.create",
+      type: "com.example.widget.delete",
+    },
+  ],
+});
+DiscoveryService.registerService({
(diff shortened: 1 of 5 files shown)


grant:
This looks like a really cool feature.

I can't find the discovery features in the SDK spec:
https://github.com/cloudevents/spec/blob/master/SDK.md

I'd imagine discovery would be a separate module than the main SDK, right?

loopingz:
It can be separated feature, I was unsure as it is pretty small and could imagine it would be useful to other people, having 2 or 3 SDK seems a bit overkill to me, but I'm not strong about it.

The spec is available here: https://github.com/cloudevents/spec/blob/master/discovery.md

lance:
@loopingz not sure if you noticed, but I submitted a PR to your fork to fix all of the TypeScript linting issues. Pushing those to this PR should help to make reviews easier.

examples/express-ex/index.js
```diff
@@ -1,22 +1,48 @@
-/* eslint-disable no-console */
+/* eslint-disable */
 
 const express = require("express");
-const {Receiver} = require("cloudevents");
-
+const { Receiver, DiscoveryService } = require("cloudevents");
```
grant, line 4: Can the discovery sample be in a separate example folder?

Folks wishing to set up a server to receive events are likely different than those who want to discover events.

grant (commented):
Thanks for the PR!
- The discovery sample should be different than the more common use-case of receiving / CloudEvents
- The diff looks large, it's hard to review. I think this is because we have both `master` and `main`, where `master` is out of date. Perhaps rebase to `main`?

examples/express-ex/index.js
loopingz: Yep i will split

loopingz:
> Thanks for the PR!
> 
I did rebase from master, the PR is old so I need to update it to reflect this change, I just forgot you will get notifications right away

loopingz:
Closing this one to reopen with main branch on my repo to avoid any confusion

Issue #237, opened 2020-07-02
Receiver.accept() should sanitize headers


lance:
Most of the time users will be using the `Receiver` in some sort of web framework that automatically sanitizes headers before user code accesses them. That's not always the case, though. Because the `Receiver.acccept()` method doesn't sanitize, it won't be able to detect a structured event if the content type header is `Content-Type` instead of `content-type`.

https://github.com/cloudevents/sdk-javascript/blob/master/src/transport/receiver.ts#L62

PR #239, opened 2020-07-03
fix: ensure that the HTTP receiver sanitizes headers in accept()

lance:
Even though the underlying structured and binary receivers already sanitize
the headers, this needs to be done at the receiver.accept() level since
the headers are inspected there to determine what mode the event is being
sent as.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/237


lance:
Ping @cloudevents/sdk-javascript-maintainers PTAL

PR #246, opened 2020-07-13
Validate extension names

### LINKED ISSUE #228: CloudEvent should only allow [^a-z|0-9$] for extension names

lholmquist:
This validates the extension names based on the spec, https://github.com/cloudevents/spec/blob/master/spec.md#attribute-naming-convention

fixes #228 

### DIFF
--- a/src/event/cloudevent.ts
+++ b/src/event/cloudevent.ts
@@ -96,6 +96,11 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
 
     // finally process any remaining properties - these are extensions
     for (const [key, value] of Object.entries(properties)) {
+      // Extension names should only allow lowercase a-z and 0-9 in the name
+      // names should not exceed 20 characters in length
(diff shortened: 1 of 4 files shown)


lholmquist:
While #229 is also related,  i'm going to split them into 2 PRs

lholmquist:
> surprised there weren't more changes required to the tests.
I guess you wrote them perfectly :)



for the examples,  yea,  we are going to need to update them.  which we might be good to do in its own PR

lance:
> for the examples, yea, we are going to need to update them. which might be good to do in its own PR

Sounds good. I've added https://github.com/cloudevents/sdk-javascript/issues/249

PR #251, opened 2020-07-15
lib: validate extension values

### LINKED ISSUE #229: CloudEvent extensions should be restricted to types allowed in spec

lholmquist:
This validates the value of the cloud event extension based on the spec, https://github.com/cloudevents/spec/blob/master/spec.md#type-system

I still need to add a check for the value being a URI

fixes #229

### DIFF
--- a/src/event/cloudevent.ts
+++ b/src/event/cloudevent.ts
@@ -9,7 +9,7 @@ import {
   CloudEventV1OptionalAttributes,
 } from "./interfaces";
 import { validateV1, validateV03 } from "./spec";
-import { ValidationError, isBinary, asBase64 } from "./validation";
+import { ValidationError, isBinary, asBase64, isValidType } from "./validation";
 import CONSTANTS from "../constants";
 import { isString } from "util";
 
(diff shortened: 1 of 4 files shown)


lance:
@lholmquist I wonder if we actually need to check for URI. There is no defined URI type in JavaScript, so the actual representation will always be a string. And I don't think we need to be parsing strings to determine if they are a URI. Maybe this can just land as is. What do you think?

lholmquist:
@lance that sounds good to me.  The only thing i could think was that we could create a new `URL` object and maybe do some parsing,  but if this really is only a string,  then maybe it doesn't make sense to get so involved.  I'll click the "Ready for Review" button then

src/event/cloudevent.ts
```diff
@@ -101,6 +101,13 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
       if (!key.match(/^[a-z0-9]{1,20}$/)) {
         throw new ValidationError("invalid extension name");
       }
+
+      // Value should be spec complient
```
lance, line 105:
```suggestion
      // Value should be spec compliant
```

Nit

lholmquist: Not picky at all :)

Issue #258, opened 2020-07-23
HTTPReceiver throws when data attribute is null or undefined


grant:
I'm seeing multiple issues when trying to using this library and testing with a `cURL` request.

These are the issues:

- `TypeError: no cloud event detected`
  - Then I added CE headers. I would assume I'd still be able to accept CloudEvents even if it was missing a header.
- `TypeError: payload is null or undefined`
  - How do you specify this? What do you mean? I tried adding a HTTP body with no luck.

I'd expect this module to simply extract the CloudEvent from a Node HTTP request and not fuss around too much! It's easier to just not use the SDK and just inspect the headers.

```js
// CloudEvents Setup
const {HTTPReceiver} = require("cloudevents-sdk");
const receiver = new HTTPReceiver();

/**
 * Parse headers and body from HTTP request.
 */
exports.extractCloudEvent = function(req) {
  return receiver.accept(req.headers, req.body);
}
```

It's easier to just not use the SDK and parse the HTTP headers manually.

grant:
Here's an example request that I would expect to work:

```sh
curl \
-H 'ce-id: projects/serverless-com-demo/logs/cloudaudit.googleapis.com%2Fdata_access-4pajkbeh8mwi159545fe7504177593' \
-H 'ce-type: com.google.cloud.auditlog.event' \
-H 'ce-specversion: 1.0' \
-H 'ce-subject: storage.googleapis.com/projects/_/buckets/test-demo_cloudbuild/objects/source/1595457496.641377-d046ef5b652b4da892a9354aa2ebfcf4.tgz' \
-H 'ce-dataschema: type.googleapis.com/google.logging.v2.LogEntry' \
-H 'ce-time: 2020-07-22T22:38:24.177593990Z' \
-H 'ce-source: //storage.googleapis.com/projects/serverless-com-demo' \
-XPOST $RUN_SERVICE_URL
```

grant:
I'm seeing error in production too with Google Cloud.

```
TypeError: payload is null or undefined
```

I assure you, we're sending folks valid CloudEvents with the product.

lance:
Looks like we are treating `data` as a required attribute. Feel free to contribute a fix as you are also a committer on the project. I will try and take a look today.

lance:
> TypeError: payload is null or undefined
> How do you specify this? What do you mean? I tried adding a HTTP body with no luck.

[...]

PR #259, opened 2020-07-23
fix: do not require an HTTP body on incoming binary messages

lance:
This commit modifies the HTTP receivers/parsers to allow for the incoming body
of an HTTP request to be empty if the event message is sent using the binary
mode. In structured mode, a `ValidationError` will still be thrown, since the
entire event must be encoded in the HTTP body.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/258


lance:
@lholmquist what do you think? Either way is fine with me.

lholmquist:
@lance we might have to wait until 3.0.0 since these changes might not merge nicely with the re-write?

PR #270, opened 2020-07-23
refactor: combine v03 and v1 event interfaces, specs and schemas into single files

grant:
Unsplits v03 and v1 as there are lots of similarities between versions of CloudEvents. It's much nicer to edit one file with two edits than two files with one edit.

Does not change functionality.

Negative PR diffs are always good! We can probably combine the interfaces at a later point too.


grant:
Pulled master.

Issue #273, opened 2020-07-24
Simplify Binary/Structured Receiver


grant:
**Describe the Bug**

I'd expect the Binary/Structured Receiver to be really easy to setup. A developer shouldn't need to know about what mode the CloudEvent is in when receiving.

These methods should be static too.

Right now there's a lot of complicated logic for the binary/structured HTTP receiver. They should probably be 1 file each.

lance:
This is why we export `Receiver` and `Emitter` now. Developers are not required to use `BinaryReceiver` or `StructuredReceiver`.

lance:
> Right now there's a lot of complicated logic for the binary/structured HTTP receiver. They should probably be 1 file each.

What specifically do you have a problem with?

grant:
> What specifically do you have a problem with?

- `Receiver` is an object, when it should be static.
- There is no generic HTTP receiver. Binary and Structured receivers should really just be combined. This is how other SDK. A consumer doesn't know if a CE is binary/structured before it is sent to the server.
  - https://github.com/cloudevents/sdk-javascript/blob/master/src/transport/http/binary_receiver.ts
  - https://github.com/cloudevents/sdk-javascript/blob/master/src/transport/http/structured_receiver.ts

Really, we should make receiving a CloudEvent easier like is done in python with a simple function:
https://github.com/cloudevents/sdk-python/tree/v1.0.0-dev#request-to-cloudevent

grant:
If we export the `BinaryHTTPReceiver` in addition to `Receiver`, we can't stop folks from using those.

We should probably get rid of them in favor of just `Receiver`.

lance:
These classes are essentially the implementation for `Receiver`. I don't want to get rid of them. And I think exposing them is not inherently wrong.

grant:
Why do we want both `BinaryHTTPReceiver` and `StructuredHTTPReceiver`?
We can just have 1 receiver that accepts both binary and structured CloudEvents. What do you think of that?

lance:
>  We can just have 1 receiver that accepts both binary and structured CloudEvents. What do you think of that?

That is `Receiver`. It exists. We don't need to make it again.

grant:
It's frustrating talking about this. There shouldn't be multiple ways of doing things.

n3wscott:
> It's frustrating talking about this. There shouldn't be multiple ways of doing things.

I do not feel this is a productive comment. 

[...]

PR #274, opened 2020-07-24
Version parsers

grant:
## Proposed Changes

- Simplify parsers
- Remove duplicate code
- Reduces file count/complexity over versioning

## Description

- Fixes Issue #273 (partially)


helio-frota:
@grant I'm not sure but my view on that is you are trying to "consolidate versions for the future" that approach is good so :+1: to that. So for what I understood when a new v.x spec come out the project has the structure to add the things without much effort or something (?) also centralizing things in versions.ts ?

In other hand @lance's comment makes sense. So my question is your PR wording not accurate meaning or something ?

anyway I'm trying to get the point

helio-frota:
>  I prefer to have code for each version in a distinct file. Not sure how others feel about this.

@lance I have no idea how many times and how often new versions will be created/added. The approach looks good to me but not sure about the benefit related with the occurrence of the future changes/additions in versions.

:+1:

grant:
The `PassThroughParser` is not stateful and can be reused. It really should be static like I believe most every other function in this library.

Is there something you'd like me to change?

lholmquist:
I don't have any strong feelings on keeping things separate.  ATM, since there are only 2 versions, it probably doesn't matter combining them.

lance:
> Is there something you'd like me to change?

Not necessarily - was just trying to have some conversation around these questions.

grant:
To be clear, I saw a lot of improvements beyond this, but I'd never finish a concrete PR if I went chasing them.

Issue #249, opened 2020-07-14
Update examples with latest changes to extensions


lance:
See: https://github.com/cloudevents/sdk-javascript/pull/246#pullrequestreview-448367574

lholmquist:
closed via #282

PR #282, opened 2020-07-29
chore: Update examples to use latest sdk changes

lholmquist:

## Proposed Changes

- This PR updates the examples to use the latest cloudevents module(3.x)

All the examples work except one curl statement in the express example that is being addressed by #284 and since we are using the `~` those changes will get picked up once a new patch version is released, automatically

## Description
- Fixes Issue #249 
- Version:

### DIFF
--- a/examples/express-ex/README.md
+++ b/examples/express-ex/README.md
@@ -67,7 +67,7 @@ curl -X POST \
      -H'ce-source:https://github.com/cloudevents/spec/pull/123' \
      -H'ce-id:45c83279-c8a1-4db6-a703-b3768db93887' \
      -H'ce-time:2019-11-06T11:17:00Z' \
-     -H'ce-my-extension:extension value' \
+     -H'ce-myextension:extension value' \
      http://localhost:3000/
 ```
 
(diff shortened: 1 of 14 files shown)


lholmquist:
I noticed something with the express example.  The last curl statement for v03 fails with an "Invalid Payload" error.  

__A Binary One with Base 64 Encoding__

```bash
curl -X POST \
     -d'@../payload/data-1.txt' \
     -H'Content-Type:application/json' \
     -H'ce-specversion:0.3' \
     -H'ce-type:com.github.pull.create' \
     -H'ce-source:https://github.com/cloudevents/spec/pull/123' \
     -H'ce-id:45c83279-c8a1-4db6-a703-b3768db93887' \
     -H'ce-time:2019-06-21T17:31:00Z' \
     -H'ce-datacontentencoding:base64' \
     http://localhost:3000/
```

If you remove that `ce-datacontentencoding` header, it will work,  but I don't think that is the right approach.   The error is happening during this section of code:  https://github.com/cloudevents/sdk-javascript/blob/master/src/event/spec.ts#L33, when testing if the data is Binary.  but at this point,  the data, at least in the example, is an object, so that check fails

@lance  thoughts?

lance:
@lholmquist interesting... I've worked out what the problem is. See https://github.com/cloudevents/sdk-javascript/issues/284. I've got a fix, but need to add tests and check behavior with the structured receiver as well.

lance:
@lholmquist ok I have pushed a PR for a fix. I don't think that this deserves a quick version bump because it only affects encoded data on v0.3 events.

Issue #284, opened 2020-07-29
BinaryHTTPReceiver should remove datacontentencoding property after decoding 


lance:
**Describe the Bug**
When the `BinaryHTTPReceiver` accepts an incoming event, one of the first things it does is check to see if the event has its data encoded, by looking for the `CloudEvent#datacontentencoding` property. If the data __is__ encoded, and the content type is `application/json`, then it will parse the data so that to the user, it ultimately is in `Object` form.

However, when validating the event, the validator checks to see if the `datacontentencoding` is set and if so, will throw a `ValidationError` if it's not a string or Buffer.

This is only relevant for v0.3 events, since `datacontentencoding` is not part of the 1.0 specification.

**Steps to Reproduce**
1. pull in the changes from https://github.com/cloudevents/sdk-javascript/pull/282
2. run the example with "A Binary One with Base 64 Encoding" from the example README
3. watch it fail

**Expected Behavior**
The event should be accepted, created and validated without error.

lance:
Note that this is also an issue for v03 structured events.

PR #285, opened 2020-07-29
fix: ensure that data encoded as base64 is parsed as an object

lance:
## Proposed Changes

Parse incoming data as JSON for an event with `datacontentencoding` set to "base64"
and a content type of application/json or application/cloudevents+json.

## Description
The 0.3 specification states that `datacontentencoding` may be set to base64.
If an incoming event arrives over HTTP with this value set, and the content type
is either application/json or application/cloudevents+json, then we should ensure
that the data is decoded and parsed, and the `datacontentencoding` property is
removed.

- Fixes: https://github.com/cloudevents/sdk-javascript/issues/284
- See: https://github.com/cloudevents/sdk-javascript/pull/282
- Version: 3.0.0 (spec version 0.3)

### DIFF
--- a/src/transport/http/binary_receiver.ts
+++ b/src/transport/http/binary_receiver.ts
@@ -84,9 +84,6 @@ export class BinaryHTTPReceiver {
       eventObj.datacontenttype === CONSTANTS.MIME_JSON &&
       eventObj.datacontentencoding === CONSTANTS.ENCODING_BASE64
     ) {
-      process.stdout.write(JSON.stringify(eventObj));
-      process.stdout.write(JSON.stringify(parsedPayload));
-
       delete eventObj.datacontentencoding;
     }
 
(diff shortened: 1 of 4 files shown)


src/transport/http/binary_receiver.ts
```diff
@@ -77,6 +77,18 @@ export class BinaryHTTPReceiver {
...
       }
     }
+    // At this point, if the datacontenttype is application/json and the datacontentencoding is base64
+    // then the data has already been decoded as a string, then parsed as JSON. We don't need to have
+    // the datacontentencoding property set - in fact, it's incorrect to do so.
+    if (
+      eventObj.datacontenttype === CONSTANTS.MIME_JSON &&
+      eventObj.datacontentencoding === CONSTANTS.ENCODING_BASE64
+    ) {
+      process.stdout.write(JSON.stringify(eventObj));
```
lholmquist, line 87: are these needed?

lance: Oh, crap no. I meant to delete these.

Issue #298, opened 2020-08-03
Easier Access to Constants for users


lholmquist:
**Is your feature request related to a problem? Please describe.**

If a user wants to get access to the constants,  they can do that by importing/requiring like this:

```
const Constants = require('cloudevents/dist/constants').default;
```

**Describe the solution you would like to see**

While this works,  it isn't as intuitive as it could be.

It would probably be nice if it was export at the top level, so we could do something like this:

```
const { Constants } = require('cloudevents');
```

PR #299, opened 2020-08-03
lib(Constants): expose constants in a more intuitive way

lholmquist:

## Proposed Changes

This change exposes the`Constants` from the "index" so a user no longer has to path into the dist folder.  I believe this was the original way to do it before the Typescript re-write

## Description
- Fixes Issue #298
- Version: 3.x

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,8 @@ import { Receiver, Mode } from "./transport/receiver";
 import { Protocol } from "./transport/protocols";
 import { Headers, headersFor } from "./transport/http/headers";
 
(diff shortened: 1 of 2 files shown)


lholmquist:
> What do you think about using Constants instead? 

Not sure what you mean

lance:
I mean export `Constants` instead of `CONSTANTS`.

lholmquist:
> I mean export Constants instead of CONSTANTS.

Yea,  i like that.  let me turn this into a draft so i can make the change

lholmquist:
I think i might just merge it like it is,  using the all caps version.

Issue #304, opened 2020-08-06
Release Procedure Document


lholmquist:
I noticed that we don't have any documentation around performing a new release.  We have a contributors guide and how to setup PRs  but nothing on doing an actual release.

With more folks having committer status and access to doing releases, we should probably have something

Issue #300, opened 2020-08-04
Release 3.1.0


lholmquist:
3.0.1 was released the 29th of August, but we also forgot to update the browser version to the new name.   The commits below include that update.   Should we release a new patch release soon with that change?  I don't think there are any commits below that need a minor bump.  I can see an argument for a7e0aa0 as a minor bump,  but I don't think it needs to be

```
git log v3.0.1..origin/master --oneline

a7e0aa0 lib: expose constants in a more intuitive way (#299)
bc81242 build(deps): bump elliptic from 6.5.2 to 6.5.3 (#290)
31c2005 refactor: no import star (#297)
48d182b fix: update browser name to cloudevents. (#292)
763838c chore: Update examples to use latest sdk changes (#282)
3fab5f2 docs: update badge name (#289)
ed9ea95 fix: ensure that data encoded as base64 is parsed as an object (#285)
e219a30 chore: Update readme with correct Receiver usage (#287)
```

lance:
> 3.0.1 was released the 29th of August

That's some powerful time machine work you've got there! :laughing:

lance:
+1 for a new release

lholmquist:
> That's some powerful time machine work you've got there! laughing

:facepalm:

lance:
@lholmquist thinking a bit more about this, I think maybe this warrants a 3.1.0 release. The change with `CONSTANTS` being exported is a breaking change, and will actually break an automated upgrade.

lholmquist:
@lance that is fine.

lholmquist:
lets make sure this makes it in too, https://github.com/cloudevents/sdk-javascript/pull/302

lholmquist:
Updating the list of commits:

```
git log v3.0.1..origin/master --oneline

a5249de chore:  fix promise tests to break the build when they fail (#305)
ad0c434 fix: Add Correct Headers to emitted Binary Event (#302)
a7e0aa0 lib: expose constants in a more intuitive way (#299)
bc81242 build(deps): bump elliptic from 6.5.2 to 6.5.3 (#290)
31c2005 refactor: no import star (#297)
48d182b fix: update browser name to cloudevents. (#292)
763838c chore: Update examples to use latest sdk changes (#282)
3fab5f2 docs: update badge name (#289)
ed9ea95 fix: ensure that data encoded as base64 is parsed as an object (#285)
e219a30 chore: Update readme with correct Receiver usage (#287)
```

[...]

PR #306, opened 2020-08-07
docs: Release Guidelines

lholmquist:
This is the start of a release guidelines for maintainers.

I think i have everything here.  There is one TODO, asking if we should have a specified time after the last "+1" from other maintainers.

Outstanding questions:

*  does the current release script also sign and adds the "signed-off by" thing 
* Should we mention something on publishing docs here
* Should we add that we should create a "release" on github

## Description
- Fixes Issue #304

### DIFF
--- a/RELEASE_GUIDELINES.md
+++ b/RELEASE_GUIDELINES.md
@@ -22,9 +22,6 @@ This output should be pasted into the issue as normal text.  This will allow Git
 
 Before a release can be finalized, other maintainers should give a +1 or a thumbs up or some other identifying mark that they are good with the changes.
 
-// TODO: should we wait a specified time after the +1's before we create
-
-
 ## Create and Publish the release
 
 Once consensus has been reached on the proposal it is time to create the release and publish it to npm.
@@ -42,6 +39,15 @@ This will update the CHANGELOG.md and create a new tag based on the version.  Th
 ```
 git push upstream master --follow-tags
 ```
+
+### Create the release on GitHub
+
+Once the release tag has been created and pushed up to Github, we should draft a new release using the Github UI, which is [located here](https://github.com/cloudevents/sdk-javascript/releases/new)
+
+* Tag Version should be the tag that was just created
+* The release title should be something like "VERSION Release"
+* And the Changelog entries for the current release should be copied/pasted into the comments
+
 
 ### Publish to npm
 
--- a/RELEASE_GUIDELINES.md
+++ b/RELEASE_GUIDELINES.md
@@ -1,0 +1,56 @@
+# Module Release Guidelines
+
+## Create a Proposal Issue
+
+To prepare for a new release, create a [new issue](https://github.com/cloudevents/sdk-javascript/issues/new?assignees=&labels=&template=feature-request.md&title=) where the title of the issue cleary reflects the version to be released.
+
+For example: "Proposal for 3.2.0 release", or something similar.  If you are not sure which version is the next version to be released, you can run `npm run release -- --dry-run` to find out what the next version will be.
+
+The body of the issue should be the commits that will be part of the release.  This can be easily accomplished by running a git log command with a defined **range**.  This range should start at the most recent version tag and end at the latest commit in the master branch.
+
+For example:
+
+```
+git log v3.0.1..upstream/master --oneline
+```
+
+This will output all the commits from the 3.0.1 tag to the latest commits in the remote upstream/master branch.
+
+This output should be pasted into the issue as normal text.  This will allow Github to magically turn all commit hashes and PR/Issues numbers to links.
+
+### Get Consensus
+
+Before a release can be finalized, other maintainers should give a +1 or a thumbs up or some other identifying mark that they are good with the changes.
+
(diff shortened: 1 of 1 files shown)


lance:
@lholmquist 
> does the current release script also sign and adds the "signed-off by" thing

No it doesn't. Good catch - I guess a step in the process should be to amend that commit

> Should we mention something on publishing docs here

The docs are automatically published on release using a [github workflow](https://github.com/cloudevents/sdk-javascript/blob/master/.github/workflows/api-docs.yaml)

> Should we add that we should create a "release" on github

Yes, probably so. GitHub considers a new tag a release, for the purposes of workflows, but I try to copy the generated section of CHANGELOG.md and paste it into the release markdown here
image

We may also want to add a step to the release process to actually create a branch for the release version, in addition to a tag. This has been discussed before and rejected, but it would make possible @grant's suggestion below.

@grant 
> Can we use a GitHub action for automatic releases on push to master?

That's a decent suggestion, but I don't think we should do that for every push to `master` since that would mean a release for every PR that lands. If we create a branch for each release, then we could have a workflow handle the actual releases, and our release process would be simply pushing the new branch. The workflow would detect the branch and then do the steps that are outlined above. But I'm not sure I love this idea.

RELEASE_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+
+This will output all the commits from the 3.0.1 tag to the latest commits in the remote upstream/master branch.
+
+This output should be pasted into the issue as normal text.  This will allow Github to magically turn all commit hashes and PR/Issues numbers to links.
+
+### Get Consensus
+
+Before a release can be finalized, other maintainers should give a +1 or a thumbs up or some other identifying mark that they are good with the changes.
+
+// TODO: should we wait a specified time after the +1's before we create
```
grant, line 25: I don't think this is necessary. All PRs should only land if a maintainer approved it.

RELEASE_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+git log v3.0.1..upstream/master --oneline
+```
+
+This will output all the commits from the 3.0.1 tag to the latest commits in the remote upstream/master branch.
+
+This output should be pasted into the issue as normal text.  This will allow Github to magically turn all commit hashes and PR/Issues numbers to links.
+
+### Get Consensus
+
+Before a release can be finalized, other maintainers should give a +1 or a thumbs up or some other identifying mark that they are good with the changes.
```
grant, line 23: I'm fine not having this step so long as all commits are approved.

RELEASE_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+
+// TODO: should we wait a specified time after the +1's before we create
+
+
+## Create and Publish the release
+
+Once consensus has been reached on the proposal it is time to create the release and publish it to npm.
+
+### Create the Release
```
grant, line 32: If we can use a release bot instead of scripts, that may make the process easier and less error prone.

RELEASE_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+npm run release
+```
+
+This will update the CHANGELOG.md and create a new tag based on the version.  This can then be pushed upstream by doing:
+
+```
+git push upstream master --follow-tags
+```
+
+### Publish to npm
```
grant, line 46: A release bot would sent a PR that would automatically publish when the PR is merged by a human.

grant:
Yup, not sure exactly how the bot would work with the branching, but I definitely recommend setting up GitHub actions for automatic release PR creation and automatic publishing when the PR is merged by a human.

RELEASE_GUIDELINES.md
lholmquist: what about 2FA on npm,  Not sure i really want to store a token anywhere

lholmquist:
> Can we use a GitHub action for automatic releases on push to master?

Similar to @lance 's thought,  i'm not a big fan of having a release on every single push to master.  We don't really need a new release everytime a new snyk PR is merged or a typo on a Readme.


I think we should revisit the "create a new branch for a release" workflow and see how github actions fits into that.  I'm also reminded of this tweet i saw the other day:

> "Never spend 6 minutes doing something by hand when you can spend 6 hours failing to automate it" :) 

The main reason for the document, is that we have a pending release, #300 , that we should probably get out this week,  and we didn't have any document about the process.  So if this works for our current process, then we can iterate on it later

lance:
> The main reason for the document, is that we have a pending release, #300 , that we should probably get out this week, and we didn't have any document about the process. So if this works for our current process, then we can iterate on it later

Definitely agree with this. Let's get through this doc using the existing processes and open an issue around release automation.

lholmquist:
addressing the signoff of the release commit in this PR https://github.com/cloudevents/sdk-javascript/pull/307

RELEASE_GUIDELINES.md
grant: For publishing with bots, I don't believe many packages enable 2FA for publishing (different than login). I think there are blogposts about this.

PR #310, opened 2020-08-11
refactor: simplify http emitter

grant:
## Proposed Changes

- Simplifies http emitter to reduce duplicated code. A larger part of making emitter simpler, such that you can simply `emit()` a CloudEvent.


grant:
Hi, I understand that there are changes to the filename that affect the API.
We used to have many index files that simply imported and exported. We have removed most of them, but this one is left. I agree, refactoring that by itself isn't very meaningful, but it's really part of a larger effort.

I created https://github.com/cloudevents/sdk-javascript/issues/314 to describe a bit of what I'm talking about.

I also added more detail to the PR description.

---

This change is required for eventually creating an easier to use emit function.

lance:
> Simplifies http emitter to reduce duplicated code
This change is not reducing duplicated code.

>  you can simply emit() a CloudEvent.
You can already simply emit() a CloudEvent.

grant:
There is duplicated code in imports. The diff count is negative.
It has benefit as the interface is simpler and easier to refactor. There are fewer exports, which is a good thing.
It does not attempt to make the emitter static.

What would you like me to change in this PR?

lance:
> What would you like me to change in this PR?

Nothing. I don't want it to land because it's a breaking change with no obvious benefit. If it's in preparation for an actual beneficial change, then submit that actual beneficial change with this included. I'm not opposed to breaking changes in the abstract, but in this case there really is no benefit at all to the end user and so I don't think it should land until there is one.

grant:
I'm purposefully creating small PRs that don't break tests. I can't create a larger PR for this. They would depend on this. (Believe me, I've tried.)

Do you disagree that removing duplicate code and duplicate exports is not a benefit?

lholmquist:
> I'm purposefully creating small PRs that don't break tests. I can't create a larger PR for this. They would depend on this. (Believe me, I've tried.)

I'm curious why you can't create a large PR that takes care of all the refactoring you feel is necessary?     I know you mentioned not wanting to break tests,  but just fix the tests then in that same PR.

When we moved to typescript,  that was a fairly large PR,  and the tests broke and were re-written to use the new APIs

grant:
Hi folks,
Can you please review the code of this PR. Please request changes with specific requests.

Fundamentally we're replacing 2 exports with 1 export for the same HTTP emitter. We've already done this in other files in different PRs. Besides the not exporting twice, the PR is a refactor.

lholmquist:
> Fundamentally we're replacing 2 exports with 1 export for the same HTTP emitter. We've already done this in other files in different PRs. Besides the not exporting twice, the PR is a refactor.


it is not as simple as that statement.   Yes, you are combining 2 exports with 1, but you are also introducing a breaking change for the user,  with not really any benefit to them as I mentioned in my previous comment.

> Do you disagree that removing duplicate code and duplicate exports is not a benefit?

I also wanted to address this statement.  This is not a yes or no answer.  it depends on the circumstances.  I would say generally yes,  less code is good,  but only when it doesn't break things for an end user.  And in this case, as said before, it is breaking things, for a small cleanup

grant:
I'm not sure what you want me to do.
Yes, many PRs in this repo have been breaking changes.

lholmquist:
> Yes, many PRs in this repo have been breaking changes.

And the PRs that have had breaking changes, have added value to the end user,  like the big Typescript re-write.


I think to land this, we would have to wait until a 4.0 release.

grant:
4.0 is fine. How can I help land this in 4.0?

lance:
> 4.0 is fine. How can I help land this in 4.0?

I recommend closing this PR and making these changes as a part of another PR that adds tangible value to the user.

grant:
Will just close.

I really really think that small PRs are okay. Eventually whoever tackles https://github.com/cloudevents/sdk-javascript/issues/314 can just take this PR and make it a bigger PR.

When I tried to do that, I ran into issues modifying too many tests, so it may take time.

PR #321, opened 2020-08-19
chore: Update README with correct links for the support specification versions

### LINKED ISSUE #320: Readme:  Supported Specification Features has broken links
**Describe the Bug**

If you were to click the v0.3 and v1.0 links on this table on the readme, https://github.com/cloudevents/sdk-javascript#supported-specification-features,  they return a 404 since they don't link anywhere.

We could/should probably link out to the appropriate section in the spec repo.

I did notice however, that the spec repo only has a master and v1.0 branches,  https://github.com/cloudevents/spec/branches/all so i'm not sure where to get the v0.3 spec links

lholmquist:
This updates the support specification headers with the correct links to their respective versions

There was also a stray `HTTPReceiver` in the readme examples that i took care of here

fixes #320

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ npm install cloudevents
 
 #### Receiving Events
 
-You can choose almost any popular web framework for port binding. Use an
-`HTTPReceiver` to process the incoming HTTP request. The receiver accepts
+You can choose almost any popular web framework for port binding. Use a
+`Receiver` to process the incoming HTTP request. The receiver accepts
 binary and structured events in either the 1.0 or 0.3 protocol formats.
 
 ```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,18 +115,18 @@ There you will find Express.js, TypeScript and Websocket examples.
 
 ## Supported specification features
 
-| Core Specification |  [v0.3](v03spec) | [v1.0](v1spec) |
+| Core Specification |  [v0.3](https://github.com/cloudevents/spec/tree/v0.3) | [v1.0](https://github.com/cloudevents/spec/tree/v1.0) |
```
lance, line 118: Can these links be to the actual spec document instead of just to the top level of the repo? The others make sense as top-level since they are about multiple specs, but this is the primary spec source.

lholmquist: sure

lholmquist: @lance updated

Issue #322, opened 2020-08-20
Adding More Transport Protocols - A Proposal


lholmquist:
**What follows is a thought dump, so sorry if it rambles a bit.**

This module is in a pretty good state with the HTTP Protocol, so i started thinking about what the next protocol that could be added.

As I thought about it,  I realized a couple things.  

This first is we probably need to make our top level `Receiver` and `Emitter` a little more generic.  At the moment,  they have things related to the HTTP Protocol in them.

This second thins is when we add a new transport protocol in, we are also potentially going to increase the dependencies for each one added.  My assumption would be that we would use a library related to that transport protocol for emitting events.  For example,  if we added MQTT then we might use the [mqtt](https://github.com/mqttjs/MQTT.js) module.

The problem,  and maybe it isn't a problem, is that a user might only want or need 1 protocol,  and if we had multiple protocols implemented,  then they would be getting more dependencies than needed.

One solution to this could be creating new modules that depend on this module as a base for each of the new Protocols.   I would assume we would just keep the HTTP protocol in the base module since, and i have no data to back this up, it might the most common to use.  So if a user wanted the MQTT protocol, they could just install that module.  However; if they wanted to use multiple protocols,  I could see this becoming a little messy.

Another solution i thought about was some sort of plugin architecture,  but i not really sure of the UX of it. with this case we would still be creating separate modules to plugin,  but there would be one "entry point" to creating new Emitters and Receivers

Again,  this was a thought dump to start a discussion on our best path forward.

lholmquist:
@cloudevents/sdk-javascript-maintainers and anyone else

lance:
@lholmquist thanks for bringing this up! As you know, I'm interested in moving ahead with MQTT protocol support in the near future.

One thing to consider is that we don't really need to support the MQTT protocol itself, we just need to provide a means for developers to [express a `CloudEvent` as an MQTT publish message](https://github.com/cloudevents/spec/blob/v1.0/mqtt-protocol-binding.md#12-relation-to-mqtt). In fact, I would say with regard to HTTP, our choice to use Axios to send over the wire may not actually be the best thing. The module could be smaller and more concise if we simply provided methods similar to this.

```js
// send a cloud event
const ce = new CloudEvent({ /** attributes */ });
const structuredHTTPMessage = HTTP.asStructured(ce);

// then it's up to the developer to actually send these using whatever HTTP module they like
axios.request( {
  method: POST,
  data: structuredHTTPMessage.body,
[...]

Issue #295, opened 2020-07-31
isCloudEvent Function


grant:
**Describe the Bug**

Is there an easy way to determine if a HTTP request is a CloudEvent or not?

It would be great if there was a `isCloudEvent` function or something similar. Maybe just documentation if we can re-use existing functions together to achieve this.

**Steps to Reproduce**

- Try to find a function to determine if a HTTP request is a valid CE

**Expected Behavior**

- Documentation in the main README for how to determine if a HTTP req is a CE

lance:
We do not have that in this module, although I can see it being useful. For a current project I am working on, we do this before trying to parse the event. 

https://github.com/boson-project/faas-js-runtime/blob/master/lib/event-handler.js#L34-L50

grant:
@lance What do you think about this logic?

https://github.com/cloudevents/sdk-javascript/blob/master/src/event/cloudevent.ts

- `CloudEvent#isBinary`
  - True if HTTP request has 4 required CE headers, otherwise false

lance:
Hmm - I don't think that it makes sense for a static `isBinary` method on `CloudEvent` because a `CloudEvent` itself is neither binary or structured. Those are just transport representations and really only make sense in the context of an incoming request over a transport protocol. 

This would make more sense as a standalone function in `cloudevents/transport/http`. The logic for an `isCloudEvent` (distinct from `isBinary`) method should be something like this:

- if the `content-type` header is set to an [event format media type](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#14-event-formats) return true (it's a structured event)
- if the `ce-id`, `ce-type`, `ce-source` and `ce-specversion` headers exist return true (it's a binary event)
- else return false

grant:
Yup, adding the method to the HTTP transport files instead of CloudEvent files makes sense.

PR #324, opened 2020-08-22
feat: introduce Message, Serializer, Deserializer and Binding interfaces

lance:
## Proposed Changes

This PR introduces the `Message`, `Serializer` and `Deserializer`, and `Binding` interfaces used to convert a `CloudEvent` into a `Message` that can be sent across a transport protocol. The first protocol implemented for this is HTTP, and some of the functionality formerly in `src/transport/http` has been simplified, reduced and/or moved to `/src/messages/http`.  

Test for V1 events are in place. Conformance tests have been modified to use these new interfaces vs. the HTTP `Receiver` class.

### Example usage
In an express.js application, simply create a `Message` using the incoming request headers and body. Then call `HTTP.toEvent(message)`.

```js
app.post("/", (req, res) => {
  const message = {
    headers: req.headers,
    body: req.body
  };

  const event = HTTP.toEvent(message);
});
```

Of course, this can be inlined as: `HTTP.toEvent({ headers: req.headers, body: req.body });`.

Sending a `Message` should also be straightforward. Convert a `CloudEvent` to a message, and send it using Axios or any other HTTP transport library that allows setting headers and body.

```js
const message = HTTP.binary(event);
axios.post({
  url: 'https://event.receiver.com',
  headers: message.headers,
  body: message.body
});
```

- Fixes: https://github.com/cloudevents/sdk-javascript/issues/322
- Fixes: https://github.com/cloudevents/sdk-javascript/issues/295
- Version: 3.2.0


src/index.ts
```diff
@@ -18,14 +18,19 @@ export {
...
+  Serializer,
+  HTTP,
   // From transport
   Emitter,
   Receiver,
-  Mode,
   Protocol,
   TransportOptions,
-  Headers,
-  headersFor,
```
lholmquist, line 28: If we are going to release this as 3.2.0, then i think we need to still export `headersFor` and not remove it until 4.0

lance: Ahh - good point...

lance:
@cloudevents/sdk-javascript-maintainers this should be ready now with both v1 and v03 of the spec being tested. PTAL

examples/express-ex/index.js
```diff
@@ -1,8 +1,8 @@
-/* eslint-disable no-console */
+/* eslint-disable */
 
 const express = require("express");
-const {Receiver} = require("cloudevents");
```
lholmquist, line 4: Should we wait until the 3.2.0 release comes out before updating this example?

lance: Hmm - I suppose. I actually didn't mean to commit that. I was just using it to test out the new interfaces. I'm good either way. I can revert this file.

src/transport/receiver.ts
```diff
@@ -36,64 +14,13 @@ export class Receiver {
    * @param {Object|JSON} body The body of the HTTP request
    * @return {CloudEvent} A new {CloudEvent} instance
    */
-  static accept(
-    headers: Headers,
-    body: string | Record<string, unknown> | CloudEventV1 | CloudEventV03 | undefined | null,
-  ): CloudEvent {
+  accept(headers: Headers, body: string | Record<string, unknown> | undefined | null): CloudEvent {
```
lholmquist, line 17: Should we have a deprecation message printout when this is called?

Is it also possible in Typescript to deprecate a function using the jsdoc annotations?

lance:
@lholmquist I can revert those tests if you think they are needed. My problem with them is that they were not really integration tests, as they were testing implementation details and not the public facing API. Coverage doesn't actually go down when these are removed!

src/transport/receiver.ts
lance: Yes - adding `@deprecated` results in doc output that looks like this:

image

lholmquist:
>  they were testing implementation details and not the public facing API.

Ok, then it is probably fine to remove them

src/transport/receiver.ts
lholmquist: Nice,  probably want add the same to the emitter?

lance:
@lholmquist I think this is probably ready to land, but I'm not sure if your comments are an implicit approval or not.

lholmquist:
oh sorry,  i think i forgot to press the approve button

Issue #325, opened 2020-08-24
Loose event validation


lance:
**Is your feature request related to a problem? Please describe.**
Currently, when creating a `CloudEvent` instance, one of the final things that happens in the constructor before freezing the object is `this.validate()` is called. The `validate()` function will throw an exception if the event is invalid. This effectively means that there is no way to create an event that's invalid. In most cases, this is probably for the best. But what happens if you have a system with a known bug that's emitting invalid events, and some JavaScript service using this module is receiving them. There's no way to actually do that without forking this repo. 

**Describe the solution you would like to see**
- Optionally, don't call `this.validate()` in the constructor based on an added `boolean` parameter (`strict: boolean = true` by default).
- Change `Detector` interface from https://github.com/cloudevents/sdk-javascript/pull/324 to use an optional parameter for loose validation.
- Change the `Deserializer` interface which converts from a `Message` to a `CloudEvent` to use an optional parameter for loose validation.

**Additional context**
This is all in the context of creating events manually, or deserializing an incoming `Message` into a `CloudEvent`. I haven't thought much about the serialization step. If a user manually creates an invalid event, and then serializes it to a `Message`, should that throw an exception? I think maybe it should, but interested in other's opinions.

lholmquist:
When i was first adding the validation to the CloudEvent constructor,  i contemplated adding what you are talking about in.   My main reason was because of how the test were written.

I can see this being a feature,  but my only concern is if a user has a service where they send cloud events, and turn off the event validation.  then another user uses that service to receive events,  how will they know that the service has the validation turned off.  They will probably get errors when trying to "accept" the event and think that maybe there is a bug in this library.

grant:
Thanks for creating the issue!

According to the SDK spec, there are various methods described. One of them is **creation** and one is **validation**.
https://github.com/cloudevents/spec/blob/master/SDK.md#technical-requirements

I would assume the [composition method](https://github.com/cloudevents/spec/blob/master/SDK.md#compose-an-event) and [validation method](https://github.com/cloudevents/spec/blob/master/SDK.md#validation) are separate and folks can use them separately as they wish for whatever reason they want.

As such, I think the validation method should not be called by default within the composition method. It should be default `false`. This is important because consumers of CloudEvent from major CE producers can expect to have a valid event and don't need this.

i.e. we don't need `new CloudEvent({...}, false)` everywhere.

PR #328, opened 2020-08-26
feat: add a constructor parameter for loose validation

lance:
## Proposed Changes

- add a second, optional, boolean parameter called `strict` to `CloudEvent` constructor, defaulting to `true`.
- add a second, optional, boolean parameter  called `strict` to `CloudEvent.cloneWith()` defaulting to `true`
- skip all validation steps if `strict` is false
- add tests for these changes 

## Description

This commit adds a second, optional boolean parameter to the `CloudEvent`
constructor. When `false` is provided, the event constructor will not
perform validation of the event properties, values and extension names.

As I am writing this, I wonder if it will be less error prone to default to `false`
and call the parameter `looseValidation`. 

- Fixes: https://github.com/cloudevents/sdk-javascript/issues/325
- Version: 3.1.0

### DIFF
--- a/src/event/cloudevent.ts
+++ b/src/event/cloudevent.ts
@@ -46,7 +46,15 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
   schemaurl?: string;
   datacontentencoding?: string;
 
-  constructor(event: CloudEventV1 | CloudEventV1Attributes | CloudEventV03 | CloudEventV03Attributes) {
+  /**
+   * Creates a new CloudEvent object with the provided properties. If there is a chance that the event
+   * properties will not conform to the CloudEvent specification, you may pass a boolean `false` as a
+   * second parameter to bypass event validation.
+   *
+   * @param {object} event the event properties
+   * @param {boolean?} strict whether to perform event validation when creating the object - default: true
+   */
+  constructor(event: CloudEventV1 | CloudEventV1Attributes | CloudEventV03 | CloudEventV03Attributes, strict = true) {
     // copy the incoming event so that we can delete properties as we go
     // everything left after we have deleted know properties becomes an extension
     const properties = { ...event };
@@ -105,20 +113,20 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
     for (const [key, value] of Object.entries(properties)) {
       // Extension names should only allow lowercase a-z and 0-9 in the name
       // names should not exceed 20 characters in length
-      if (!key.match(/^[a-z0-9]{1,20}$/)) {
+      if (!key.match(/^[a-z0-9]{1,20}$/) && strict) {
         throw new ValidationError("invalid extension name");
       }
 
       // Value should be spec compliant
       // https://github.com/cloudevents/spec/blob/master/spec.md#type-system
-      if (!isValidType(value)) {
+      if (!isValidType(value) && strict) {
         throw new ValidationError("invalid extension value");
       }
 
       this[key] = value;
     }
 
-    this.validate();
+    strict ? this.validate() : undefined;
 
     Object.freeze(this);
   }
@@ -193,6 +201,7 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
   /**
    * Clone a CloudEvent with new/update attributes
    * @param {object} options attributes to augment the CloudEvent with
+   * @param {boolean} strict whether or not to use strict validation when cloning (default: true)
    * @throws if the CloudEvent does not conform to the schema
    * @return {CloudEvent} returns a new CloudEvent
    */
@@ -204,7 +213,8 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
       | CloudEventV03
       | CloudEventV03Attributes
       | CloudEventV03OptionalAttributes,
+    strict = true,
   ): CloudEvent {
-    return new CloudEvent(Object.assign({}, this.toJSON(), options) as CloudEvent);
+    return new CloudEvent(Object.assign({}, this.toJSON(), options) as CloudEvent, strict);
   }
 }
--- a/src/event/validation.ts
+++ b/src/event/validation.ts
@@ -8,7 +8,18 @@ export class ValidationError extends TypeError {
   errors?: string[] | ErrorObject[] | null;
 
   constructor(message: string, errors?: string[] | ErrorObject[] | null) {
-    super(message);
+    const messageString =
+      errors instanceof Array
+        ? // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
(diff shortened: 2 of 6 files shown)


src/event/cloudevent.ts
```diff
@@ -46,7 +46,15 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
...
-  constructor(event: CloudEventV1 | CloudEventV1Attributes | CloudEventV03 | CloudEventV03Attributes) {
+  /**
+   * Creates a new CloudEvent object with the provided properties. If there is a chance that the event
+   * properties will not conform to the CloudEvent specification, you may pass a boolean `false` as a
+   * second parameter to bypass event validation.
+   *
+   * @param {object} event the event properties
+   * @param {boolean?} strict whether to perform event validation when creating the object - default: true
+   */
+  constructor(event: CloudEventV1 | CloudEventV1Attributes | CloudEventV03 | CloudEventV03Attributes, strict = true) {
```
grant, line 57: Can we have strict mode default to `false` like is done with web JavaScript?

lholmquist:
> As I am writing this, I wonder if it will be less error prone to default to false and call the parameter looseValidation.

Or maybe it is just called `validate`?  what ever it is called,  Validation should happen in the default state and an action should be needed to turn it off

src/event/cloudevent.ts
lholmquist: >  like is done with web JavaScript?

What is this referring too?

grant: JavaScript is usually opt-in rather than opt-out for things like validation.

I was referring to only being strict when requested like in strict mode:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

grant:
I'd rather not have implicit validation unless explicitly requested by the user.
It would be great if a user could just call a method to validate a CE at any time if they want.

src/event/cloudevent.ts
lholmquist: not sure that really applies here,  since we aren't talking about the language level strict mode.

lholmquist:
> I'd rather not have implicit validation unless explicitly requested by the user.

Then what's the point of having a module that is trying to adhere to a spec if it is off by default?

grant:
> > I'd rather not have implicit validation unless explicitly requested by the user.
> 
> Then what's the point of having a module that is trying to adhere to a spec if it is off by default?

I would expect a separate method for validation. Or at least be optional.
I wouldn't expect it to be coupled with event creation.

https://github.com/cloudevents/spec/blob/master/SDK.md#compose-an-event

A common way to do this would be:

```ts
const ce = new CloudEvent({...});
const isValid = ce.isValid() // true or false
if (!isValid) { 
  console.log('Not a valid CloudEvent');
}
// or...
try {
  ce.validate();
} catch (e) {
  console.error(e);
}
```

I would not imagine myself doing:

```ts
let ce;
try {
  ce = new CloudEvent({...}, true);
} catch (e) {
  console.log(e)
}
```

---

There's still lots of reason to use the module like calling calling `ce.validate()`, `ce.toJSON()`, etc. Hope the above makes sense.

src/event/cloudevent.ts
grant: We're talking about opt-in validation with JS (like JS strict mode).

This PR is opt-out validation, where it could be opt-in validation.

lance:
> Then what's the point of having a module that is trying to adhere to a spec if it is off by default?

I tend to agree with you in principle here @lholmquist.

That said, a close reading of the spec makes me think that it's acceptable to default to not be strict. Specifically, https://github.com/cloudevents/spec/blob/master/SDK.md#validation only mentions that an SDK  must make it possible to validate a given Event per a spec. This implies that having invalid events is a possibility, and potentially expected. When I think about possible friction that a developer might experience when starting to work with this SDK, I can see how easy it might be to get confused or frustrated if you're testing things with an invalid event and just don't know it - and strict validation is on by default, but there's no obvious way to know this.

In spite of my tendency to prefer strict adherence to the spec always, I think maybe in this case, making the easy path as easy as possible, while providing for a simple way to remain strict could be a good thing. I don't like what it means for our tests though. By defaulting to loose validation, I think most if not all of our existing tests would need to be modified.

Maybe we can get a discussion going in the #cloudeventssdk slack to see what others think.

lholmquist:
> Maybe we can get a discussion going in the #cloudeventssdk slack to see what others think.

That sounds reasonable

grant:
Yeah, its not a clear-cut problem. I don't know what the spec says about passing invalid CEs (they should probably work unless there is a fatal issue).

Checking others with slack sgtm.

Another rationale for separate validation is because we may have already validated the CloudEvent (maybe from a different sdk, when cloning, already validated in the server, etc).

lance:
OK - I brought this issue to attention at the CE community Zoom call today and got some good feedback. Here's a proposal.

- Creating a `CloudEvent` manually using the constructor, e.g. `const ce = new CloudEvent({});` should be strict mode
- Receiving a `Message` over HTTP or other transport and converting to a `CloudEvent` via `HTTP.toEvent()` should be loose mode with no explicit validation
- Creating a `Message` from a `CloudEvent` via `HTTP.binary(event)` or `HTTP.structured(event)` should validate before creating the `Message` instance

This accomplishes a few things

- Users have to explicitly intend to create a bad `CloudEvent` manually (note: I think we should improve our error messages in this case, so that it's clear when an exception is being thrown, why exactly something went wrong)
- Allow receipt of events from external systems that may be generating bad/non-conformant events
- Prevent a user from sending bad events using this SDK, ensuring that events coming from systems using this SDK are interoperable with other conformant systems

grant:
Thanks for investigating :) Sounds fine.

My opinion is a little different (validation should be explicitly requested by the user, not done in certain cases), but the above proposal sounds fine so long as folks don't get unexpected errors.

lance:
@cloudevents/sdk-javascript-maintainers I have made the changes noted in the comments. You can see the `Message` changes tested here: https://github.com/cloudevents/sdk-javascript/pull/328/files#diff-8b36d95f2fcd988540ae11f4206992cfR53-R87 and the `CloudEvent` changes tested here https://github.com/cloudevents/sdk-javascript/pull/328/files#diff-9f6cd81066208bf471aae16589fa3389R24-R38. I have also modified the `ValidationError` class to provide much more information in the error message when an event instance is invalid https://github.com/cloudevents/sdk-javascript/pull/328/files#diff-43312dc1cddf7a559d27b62f8d838c03R11-R22. PTAL

lholmquist (approved):

grant (approved):

lholmquist:
@lance I think this is ready to be merged?

Issue #343, opened 2020-09-28
Binary data in binary format is not binary?


evankanderson:
**Describe the Bug**
I'm trying to send a cloudevent with `datacontenttype` of `image/jpeg` and the actual payload being the bytes of the image. It seems like the current code in `asData` will always base-64 the content when fetching data. (Possibly twice, if I read the code at line 12 of `http/index.js` correctly -- once for `event.data`, and once for `asData`.)

**Steps to Reproduce**
I'd like the following code to work, but it appears that `ArrayBuffer` and `Blob` are not supported types for data:

```javascript
  fetch(mediaUrl).then(async (resp) => {
    let newEvent = event.cloneWith({
      source: sourceUrl,
      type: responseEventType,
      datacontenttype: resp.headers.get("Content-Type")
    })
    // I want this. This definitely does not work, and an empty object gets encoded.
    newEvent.data = await resp.arrayBuffer();  // or .blob()
    const response = HTTP.binary(newEvent);
    res.status(200).set(response.headers).send(response.body);
  });
```

If I try to convert the arrayBuffer to a Uint32Array (`newEvent.data = new Uint32Array(await resp.arrayBuffer());`), I get the following exception:

> TypeError: Cannot assign to read only property 'data_base64' of object '[object Object]'

**Expected Behavior**
I'd like the above code to work, or to be able to find an example of sending a binary file (i.e. an image or a proto or something) as an HTTP CloudEvent.

**Additional context**
I am terrible at Javascript, so feel free to point out I'm holding it wrong.

lance:
Hi @evankanderson can you please try the `data` assignment in the `cloneWith()` method?

```js
let newEvent = event.cloneWith({
      source: sourceUrl,
      type: responseEventType,
      datacontenttype: resp.headers.get("Content-Type"),
      data: new Uint32Array(await resp.arrayBuffer())
    });
```
This will get around the assignment errors. But you are correct that when serializing an event, it will be encoded as base64. This is based on a reading of the [JSON format spec](https://github.com/cloudevents/spec/blob/v1.0/json-format.md#31-handling-of-data).

> If the implementation determines that the type of data is Binary, the value MUST be represented as a JSON string expression containing the Base64 encoded binary value, and use the member name data_base64 to store it inside the JSON object.

The assumption made in this module is that when serializing for HTTP, the event is converted into a JSON representation, and that data should be encoded as base64. But your use case clearly shows that this should not happen.

[...]

PR #344, opened 2020-09-28
fix: do not alter an event's data attribute

lance:
## Proposed Changes

- Change `CloudEvent` so that it never modifies the `data` attribute
- If an event's `datacontenttype` property is a content type that we don't support, just set the data attribute - don't throw

## Description

When setting an event's data attribute we were trying to be really clever and this is problematic. Instead, keep the data attribute unchanged. Per the 1.0 specification, the data attribute is still inspected to determine if it is binary data, and if so, a data_base64 attribute is added with the contents of the data property encoded as base64.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/343


grant (approved):

lholmquist (approved):

lance:
Do not merge. I have found a bug in this implementation that I need to fix.

lance:
@cloudevents/sdk-javascript-maintainers there is an interesting discussion in Slack around this question of binary data: https://cloud-native.slack.com/archives/C9DB5ABAA/p1601416717013400

lance:
@cloudevents/sdk-javascript-maintainers PTAL - thanks

grant (approved):
Not modifying the data is great.
I think we'll talk about base64 data a bit more in the weekly SDK call too.

lholmquist (approved):

lance:
Based on the discussion in today's CNCF call, I am going to remove this test and the associated functionality.

https://github.com/cloudevents/sdk-javascript/blob/5255a19ea458ab6bd9854a42fbc83fe59b2a783f/test/integration/message_test.ts#L190-L195

Issue #340, opened 2020-09-08
Typescript incompatible type in 3.1.0 and Express 4.x


aabrams:
Trying to add in Express 4.x middleware: 

import { Receiver, CloudEvent } from "cloudevents";

	const receivedEvent: CloudEvent = Receiver.accept(req.headers, req.body);

It fails with:

(property) IncomingMessage.headers: IncomingHttpHeaders
Argument of type 'IncomingHttpHeaders' is not assignable to parameter of type 'Headers'.
  Index signatures are incompatible.
    Type 'string | string[]' is not assignable to type 'string'.
      Type 'string[]' is not assignable to type 'string'.ts(2345)

lance:
It appears that the `IncomingHttpHeaders` type is from [Node core](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/ts3.1/http.d.ts#L7), so I think this will probably be a common issue. We should consider modifying our type.

https://github.com/cloudevents/sdk-javascript/blob/1fa3a05aed84bb4bdb225b3e2e11aba60d80efe8/src/message/index.ts#L21-L23

In the meantime @aabrams can you try

```js
const receivedEvent: CloudEvent = Receiver.accept(req.headers as Headers, req.body);
``

aabrams:
Trying what you suggested (req.headers as Headers):

Argument of type 'Headers' is not assignable to parameter of type 'import("c:/Users/adam_a/repos/Segment-Webhook/node_modules/cloudevents/dist/transport/http/headers").Headers'.
  Index signature is missing in type 'Headers'.ts(2345)
Conversion of type 'IncomingHttpHeaders' to type 'Headers' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.ts(2352)
Conversion of type 'IncomingHttpHeaders' to type 'Headers' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'IncomingHttpHeaders' is missing the following properties from type 'Headers': append, delete, get, has, and 6 more.ts(2352)

PR #346, opened 2020-10-01
fix: extend Node.js IncomingHttpHeaders in our Headers type

lance:
## Proposed Changes

- extend Node.js `IncomingHttpHeaders` in our `Headers` type

## Description

Changes the `Headers` type to make it more compatible with Node.js TypeScript projects.

- Fixes: https://github.com/cloudevents/sdk-javascript/issues/340
- Version:

### DIFF
--- a/src/message/http/index.ts
+++ b/src/message/http/index.ts
@@ -71,7 +71,7 @@ export function deserialize(message: Message): CloudEvent {
  * @returns {Mode} the transport mode
  */
 function getMode(headers: Headers): Mode {
-  const contentType = headers[CONSTANTS.HEADER_CONTENT_TYPE];
+  const contentType = headers[CONSTANTS.HEADER_CONTENT_TYPE] as string;
   if (contentType && contentType.startsWith(CONSTANTS.MIME_CE)) {
     return Mode.STRUCTURED;
   }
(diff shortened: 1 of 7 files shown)


src/message/http/index.ts
```diff
@@ -71,7 +71,7 @@ export function deserialize(message: Message): CloudEvent {
  * @returns {Mode} the transport mode
  */
 function getMode(headers: Headers): Mode {
-  const contentType = headers[CONSTANTS.HEADER_CONTENT_TYPE];
+  const contentType = headers[CONSTANTS.HEADER_CONTENT_TYPE] as string;
```
aldirrix, line 74: i think you can add as const to the actual constants file instead of in every usage of this.
Ref: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

in `sdk-javascript/src/constants.ts`
```js
const CONSTANTS = Object.freeze({
  CHARSET_DEFAULT: "utf-8",
  ... 
} } as const);

export default CONSTANTS;

File: test/integration/message_test.ts [hunk 2]
```diff
@@ -85,6 +86,25 @@ describe("HTTP transport", () => {
     }).to.throw;
   });
 
+  it("Can be created with Node's IncomingHttpHeaders", () => {
```
aldirrix, line 89: nice!

test/integration/emitter_factory_test.ts
```diff
@@ -48,7 +48,7 @@ function superagentEmitter(message: Message, options?: Options): Promise<unknown
   }
   // set headers
   for (const key of Object.getOwnPropertyNames(message.headers)) {
-    post.set(key, message.headers[key]);
+    post.set(key, message.headers[key] as string);
```
aldirrix, line 51: i guess we also have to submit an issue to superagent 😂

lholmquist (approved):

Issue #333, opened 2020-09-02
Avoid doing console.error


satazor:
This is not really a bug, nor a feature request, so I haven't followed any of the templates.

It seems `console.error` is being used in several situations, e.g.: https://github.com/cloudevents/sdk-javascript/blob/6cd310c14168013a4298f87fd8ec3ef51782d7ce/src/event/cloudevent.ts#L170.

In most scenarios, these `console.error` calls are unnecessary as errors will be thrown anyway, that should be handled by devs consuming this library. In other situations, errors aren't thrown, and these messages are useful. However, most devs are using custom loggers (like JSON loggers), and would like to be able to customize the logging behavior.

lance:
Thanks @satazor for opening this issue. It actually looks like there are only two calls to `console.error()` - the one you pointed out, and this one

https://github.com/cloudevents/sdk-javascript/blob/6cd310c14168013a4298f87fd8ec3ef51782d7ce/src/message/http/index.ts#L51

I think probably the one you note is the only one that should be removed, as the one in `src/messages/http` is informational and an exception isn't thrown.

satazor:
@lance 👍

For informal logs, please consider adding the ability to specify the log interface (maybe the same interface as console), so that we can override with custom loggers.

PR #349, opened 2020-10-07
src(receiver): don't default specversion on receiving if there isn't one.

### LINKED ISSUE #332: Should throw if specversion attribute is missing
**Describe the Bug**

Does not throw if event doesn't have a `specversion` attribute, and fallbacks to `1.0`. The spec says that the `specversion` attribute is mandatory, so I think this library should fail when it's missing, instead of "magically" falling back.

**Steps to Reproduce**
1. `Receiver.accept({ 'Content-Type': 'application/cloudevents+json' }, { id: 'foo', source: 'foo', type: 'foo' });`
2. Prints `console.error` saying it will fallback to `1.0`
3. Validation passes

**Expected Behavior**
It should throw, since the `specversion` attribute is missing

lholmquist:
* When receiving an event,  we should not default the specversion to 1.0 if no specverion field is detected.
* If no specversion is detected or the specversion does not conform to a currently supported version, we will now throw a ValidationError

* This is a BREAKING_CHANGE

fixes #332, #333

### DIFF
--- a/src/message/http/index.ts
+++ b/src/message/http/index.ts
@@ -54,10 +54,9 @@ export function isEvent(message: Message): boolean {
 export function deserialize(message: Message): CloudEvent {
   const cleanHeaders: Headers = sanitize(message.headers);
   const mode: Mode = getMode(cleanHeaders);
-  let version = getVersion(mode, cleanHeaders, message.body);
+  const version = getVersion(mode, cleanHeaders, message.body);
   if (version !== Version.V03 && version !== Version.V1) {
-    console.error(`Unknown spec version ${version}. Default to ${Version.V1}`);
-    version = Version.V1;
+    throw new ValidationError(`Unknown spec version: ${version}`);
   }
   switch (mode) {
     case Mode.BINARY:
@@ -105,7 +104,7 @@ function getVersion(mode: Mode, headers: Headers, body: string | Record<string, 
     // structured mode - the version is in the body
     return typeof body === "string" ? JSON.parse(body).specversion : (body as CloudEvent).specversion;
   }
-  return Version.V1;
+  // return Version.V1;
 }
 
 /**
--- a/test/integration/http_receiver_test.ts
+++ b/test/integration/http_receiver_test.ts
@@ -24,6 +24,32 @@ describe("HTTP Transport Binding Receiver for CloudEvents", () => {
       expect(Receiver.accept.bind(Receiver, {}, payload)).to.throw(ValidationError, "no cloud event detected");
     });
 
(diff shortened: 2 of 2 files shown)


lholmquist:
Did we need to id also?

lance (commented):
@lholmquist thinking about this... when we were discussion validation not too long ago, the point was to have loose validation on receipt of a cloud event (since you could be receiving an event from a "faulty" system). But we could give the user to explicitly validate afterwards with the `validate()` function. I think probably a misspoke when I said we should throw (sorry!). 

The goal would be to create an event without the specversion (or ID) set, and allow for user validation to deal with errors. The only problem is how to do this without changing the existing behavior for users who just want to create an event. Ideally they really shouldn't be required to specify a version - it should just default to 1.0.

lholmquist:
let me read that discussion, https://github.com/cloudevents/sdk-javascript/pull/328#issuecomment-682059664, and get back to you

lholmquist:
Just want to the workflow straight in my head.   A new event/message comes in and is received.  The specversion is not something this sdk know about or is missing.  No error happens and a new cloudevent is created.  The user then calls the CE validate function and receives an error because of the non-valid specversion field,  they handle it.   


I think the only thing we should change is that we don't default the value, since with the loose validation,  it is working as intended.  And if a user needs to handle the error,  I think they can just use the CE.cloneWith method?

lholmquist:
> I'm kind of torn on this... Requiring a user to specify the spec version when creating a new event seems a little unfortunate, since it will pretty much always be Version.V1.

I think we should still default when using the `new CloudEvent()` constructor.  There error would only happen if a user runs validate when receiving an Event from somewhere, at least that is how i'm understanding the issue.   So,  is there really an issue with us defaulting to v1?

lholmquist:
@lance i wondering what the next steps are here.  This PR is probably not correct the way it is written after reading some of our discussions,  i wonder if there really is an issue at all?

lance:
@lholmquist sorry - I have been heads down on an upcoming release and have not had much time to think about this issue. Honestly, I think the way it is now is __ok__ but not great. However, I'm not opposed to just letting this sit and stew for a while.

lholmquist:
> Honestly, I think the way it is now is ok but not great. However, I'm not opposed to just letting this sit and stew for a while.

Ok,  i'm going to close this PR, since it is not the correct way of doing it, if we thought about doing it.   Are we still blocked by this for the 4.0 release then?

lance:
> Are we still blocked by this for the 4.0 release then?

I think we are probably good to release next week.

PR #352, opened 2020-10-15
chore: tag v3.2.0 as release-v3.2.0 for release-please

lance:
## Description

This is an empty commit using a `release-please` branch name. The hope is that this will fix `release-please` per [this suggestion](https://github.com/GoogleCloudPlatform/release-please-action/issues/78).


bcoe:
@lance I think this might still cause some issues as release-please is configured to look at your main branch currently, not `cloudevents:v3.x`, if the goal is to use release-please only on the `3.x` branch, I think we'll need to expose a configuration option to support the alternate branch (_this should be possible now._).

To get releases working properly on the 2.x release line, we need a similar commit to this -- then I think we'll be on the right track.

lance:
> To get releases working properly on the 2.x release line, we need a similar commit to this -- then I think we'll be on the right track.

@bcoe thanks for your help with this. I guess it's kind of confusing. We branched for the 3.x line because we were making incompatible changes on `main`. I think I can fix this by creating the release PR to point at `main` at the SHA where we branched.

lance:
Closing and will try again...

bcoe:
@lance the historical logic looked at the highest SemVer range tag, this would fail because the tag for `3.x.x` is the largest tag, but points to a history that has diverged from your main branch.

The new logic which we're just starting to test, instead looks for a branch of the form `release-vx.x.x` -- so, the higher SemVer tag should no longer cause the logic to fail, once one of these branches exists on the `main` release line.

The problem is that this repo already existed, and has no such release branches. Basically, we need a branch of this form on whatever branch you want to release from -- in theory, this new logic should eventually allow multiple release lines at the same time (_but this functionality isn't fully baked).

PR #362, opened 2020-11-07
Remove All API's that are labeled "Remove in 4.0"

lholmquist:
* This removes all the APIs that were deprecated in the 3.x releases and marked as "remove in 4.0".

* Also removes any tests associated with those API's

Should be merged before 4.0 release.   I'll add this issue to the 4.0 proposal


lance (approved):
Sorry for the latency! This lgtm in general, but depending on which lands first, either you or @loopingz will need to adjust, since `Emitter` is sticking around as a type.

grant (approved):

PR #365, opened 2020-11-12
feat: add other interfaces from the spec for discovery

loopingz:
Signed-off-by: Remi Cattiau <remi@cattiau.com>

## Proposed Changes

- Add the interface from the Discovery spec
- Another package including Discovery service and Subscription service will be created
- Clean formatting of `examples/express-ex/index.js` example

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 import { CloudEvent, Version } from "./event/cloudevent";
 import { ValidationError } from "./event/validation";
-import { CloudEventV03, CloudEventV03Attributes, CloudEventV1, CloudEventV1Attributes } from "./event/interfaces";
+export * from "./event/interfaces";
 
 import {
   Emitter,
@@ -19,10 +19,6 @@ import CONSTANTS from "./constants";
 export {
   // From event
   CloudEvent,
-  CloudEventV03,
-  CloudEventV03Attributes,
-  CloudEventV1,
-  CloudEventV1Attributes,
   Version,
   ValidationError,
   // From message
--- a/src/event/interfaces.ts
+++ b/src/event/interfaces.ts
@@ -271,3 +271,172 @@ export interface CloudEventV03OptionalAttributes {
    */
   [key: string]: unknown;
 }
+
+/**
+ * Definition of a CloudEvents Service
+ */
+export interface CloudEventV1Service {
+  /**
+   * [REQUIRED] A URL that references this Service. This value MUST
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -1,6 +1,6 @@
 import { CloudEvent, Version } from "./event/cloudevent";
 import { ValidationError } from "./event/validation";
-import { CloudEventV03, CloudEventV03Attributes, CloudEventV1, CloudEventV1Attributes } from "./event/interfaces";
+export * from "./event/interfaces";
```
grant, line 3: No star imports please.

examples/express-ex/index.js
```diff
@@ -3,8 +3,8 @@
 const express = require("express");
 const { Receiver } = require("cloudevents");
 const app = express();
-const bodyParser = require('body-parser')
-app.use(bodyParser.json())
+const bodyParser = require("body-parser");
```
grant, line 6: I would not expect this change in this PR.

Was this fixed with some linting tool? Perhaps we should add it as a GitHub action step to check this folder.

Can we revert this file?

grant (approved):
A few requested changes. O.w. lgtm.

src/index.ts
loopingz: I think it makes more sense to export all interfaces directly, what is the reason behind no star in this case?

examples/express-ex/index.js
loopingz: Yep

lance (commented):
LGTM, though I'm still unsure whether these should live here or in a future `cloudevents-discovery` module created in this repo.

loopingz:
It could be both but Discovery and Subscription might have a common definition, that is why I think SDK holding the definition might be a good way, as it is defining the common vocabulary (interfaces).
But at the same time, they might evolve more, it is a broader issue is cloudevents spec is the cloudevent/discovery/subscription or just cloudevent.

lance:
Hmm yes. If the Subscription API and Discovery API each make use of types from the other, then I agree it makes sense for those types to live here.

src/index.ts
grant: Yeah, this looks fine actually!

grant (changes requested):
It looks like this PR is out of date.

I suggest at this moment we encourage development of discovery features in a separate discovery-centric repo.

Removing approval since this repo has changed. Ping for re-review. :)

PR #366, opened 2020-11-12
Emitter

loopingz:
## Proposed Changes

- Update Emitter from deprecated to singleton paradigm
- Allow implementing EventEmitter 
- It makes it cleaner as from any point in the source code you can just use `.emit()` of a CloudEvent to send it, you do not care about how it is sent to the right target

We might want to consider removing the EventEmitter and adding the listeners' management to be able to ensure async methods so we can emit and waiting for the event to be completely emitted before continue processing. (my favorite options will discuss it tomorrow during our call)

I was not able to avoid the warning for the `<any>` in the unit test, usage of `unknown` in the sdk is not helping there

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,6 +30,7 @@
     }],
     "valid-jsdoc": "warn",
     "semi": ["error", "always"],
-    "quotes": ["error", "double", { "allowTemplateLiterals": true }]
+    "quotes": ["error", "double", { "allowTemplateLiterals": true }],
+    "@typescript-eslint/no-explicit-any": "off"
   }
 }
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ function sendWithAxios(message) {
 
 const emit = emitterFor(sendWithAxios, { mode: Mode.BINARY });
 // Set the emit
-Emitter.getSingleton().on("event", emit);
+Emitter.on("cloudevent", emit);
 
 ...
 // In any part of the code will send the event
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # JavaScript SDK for CloudEvents
 
-[![Codacy Badge](https://api.codacy.com/project/badge/Grade/bd66e7c52002481993cd6d610534b0f7)](https://www.codacy.com/app/fabiojose/sdk-javascript?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=cloudevents/sdk-javascript&amp;utm_campaign=Badge_Grade)
-[![Codacy Badge](https://api.codacy.com/project/badge/Coverage/bd66e7c52002481993cd6d610534b0f7)](https://www.codacy.com/app/fabiojose/sdk-javascript?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=cloudevents/sdk-javascript&amp;utm_campaign=Badge_Coverage)
+[![Codacy Badge](https://api.codacy.com/project/badge/Grade/bd66e7c52002481993cd6d610534b0f7)](https://www.codacy.com/app/fabiojose/sdk-javascript?utm_source=github.com&utm_medium=referral&utm_content=cloudevents/sdk-javascript&utm_campaign=Badge_Grade)
+[![Codacy Badge](https://api.codacy.com/project/badge/Coverage/bd66e7c52002481993cd6d610534b0f7)](https://www.codacy.com/app/fabiojose/sdk-javascript?utm_source=github.com&utm_medium=referral&utm_content=cloudevents/sdk-javascript&utm_campaign=Badge_Coverage)
 ![Node.js CI](https://github.com/cloudevents/sdk-javascript/workflows/Node.js%20CI/badge.svg)
 [![npm version](https://img.shields.io/npm/v/cloudevents.svg)](https://www.npmjs.com/package/cloudevents)
 [![vulnerabilities](https://snyk.io/test/github/cloudevents/sdk-javascript/badge.svg)](https://snyk.io/test/github/cloudevents/sdk-javascript)
@@ -10,9 +10,9 @@ The CloudEvents SDK for JavaScript.
 
 ## Features
 
-* Represent CloudEvents in memory
-* Serialize and deserialize CloudEvents in different [event formats](https://github.com/cloudevents/spec/blob/v1.0/spec.md#event-format).
-* Send and recieve CloudEvents with via different [protocol bindings](https://github.com/cloudevents/spec/blob/v1.0/spec.md#protocol-binding).
+- Represent CloudEvents in memory
+- Serialize and deserialize CloudEvents in different [event formats](https://github.com/cloudevents/spec/blob/v1.0/spec.md#event-format).
+- Send and recieve CloudEvents with via different [protocol bindings](https://github.com/cloudevents/spec/blob/v1.0/spec.md#protocol-binding).
 
 _Note:_ Supports CloudEvent versions 0.3, 1.0
 
@@ -51,16 +51,15 @@ using the `HTTP` binding to create a `Message` which has properties
 for `headers` and `body`.
 
 ```js
-const axios = require('axios').default;
+const axios = require("axios").default;
 const { HTTP } = require("cloudevents");
 
-
-const ce = new CloudEvent({ type, source, data })
+const ce = new CloudEvent({ type, source, data });
 const message = HTTP.binary(ce); // Or HTTP.structured(ce)
 
 axios({
-  method: 'post',
-  url: '...',
+  method: "post",
+  url: "...",
   data: message.body,
   headers: message.headers,
 });
@@ -69,7 +68,7 @@ axios({
 You may also use the `emitterFor()` function as a convenience.
 
 ```js
-const axios = require('axios').default;
+const axios = require("axios").default;
 const { emitterFor, Mode } = require("cloudevents");
 
 function sendWithAxios(message) {
@@ -77,8 +76,8 @@ function sendWithAxios(message) {
   // and body in this function, then send the
   // event
   axios({
-    method: 'post',
-    url: '...',
+    method: "post",
+    url: "...",
     data: message.body,
     headers: message.headers,
   });
@@ -88,9 +87,38 @@ const emit = emitterFor(sendWithAxios, { mode: Mode.BINARY });
 emit(new CloudEvent({ type, source, data }));
 ```
 
+You may also use the `Emitter` singleton
+
+```js
+const axios = require("axios").default;
+const { emitterFor, Mode, CloudEvent, Emitter } = require("cloudevents");
+
+function sendWithAxios(message) {
+  // Do what you need with the message headers
+  // and body in this function, then send the
+  // event
+  axios({
+    method: "post",
+    url: "...",
+    data: message.body,
+    headers: message.headers,
+  });
+}
+
+const emit = emitterFor(sendWithAxios, { mode: Mode.BINARY });
+// Set the emit
+Emitter.getSingleton().on("event", emit);
+
+...
+// In any part of the code will send the event
+new CloudEvent({ type, source, data }).emit();
+
+// You can also have several listener to send the event to several endpoint
+```
+
 ## CloudEvent Objects
 
-All created `CloudEvent` objects are read-only.  If you need to update a property or add a new extension to an existing cloud event object, you can use the `cloneWith` method.  This will return a new `CloudEvent` with any update or new properties.  For example:
+All created `CloudEvent` objects are read-only. If you need to update a property or add a new extension to an existing cloud event object, you can use the `cloneWith` method. This will return a new `CloudEvent` with any update or new properties. For example:
 
 ```js
 const {
@@ -112,24 +140,26 @@ There you will find Express.js, TypeScript and Websocket examples.
 
 ## Supported specification features
 
-| Core Specification |  [v0.3](https://github.com/cloudevents/spec/blob/v0.3/spec.md) | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/spec.md) |
-| ----------------------------- | --- | --- |
-| CloudEvents Core              | :heavy_check_mark: | :heavy_check_mark: |
+| Core Specification | [v0.3](https://github.com/cloudevents/spec/blob/v0.3/spec.md) | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/spec.md) |
+| ------------------ | ------------------------------------------------------------- | ------------------------------------------------------------- |
+| CloudEvents Core   | :heavy_check_mark:                                            | :heavy_check_mark:                                            |
+
 ---
 
-| Event Formats |  [v0.3](https://github.com/cloudevents/spec/tree/v0.3) | [v1.0](https://github.com/cloudevents/spec/tree/v1.0) |
-| ----------------------------- | --- | --- |
-| AVRO Event Format             | :x: | :x: |
-| JSON Event Format             | :heavy_check_mark: | :heavy_check_mark: |
+| Event Formats     | [v0.3](https://github.com/cloudevents/spec/tree/v0.3) | [v1.0](https://github.com/cloudevents/spec/tree/v1.0) |
+| ----------------- | ----------------------------------------------------- | ----------------------------------------------------- |
+| AVRO Event Format | :x:                                                   | :x:                                                   |
+| JSON Event Format | :heavy_check_mark:                                    | :heavy_check_mark:                                    |
+
 ---
 
-| Transport Protocols |  [v0.3](https://github.com/cloudevents/spec/tree/v0.3) | [v1.0](https://github.com/cloudevents/spec/tree/v1.0) |
-| ----------------------------- | --- | --- |
-| AMQP Protocol Binding         | :x: | :x: |
-| HTTP Protocol Binding         | :heavy_check_mark: | :heavy_check_mark: |
-| Kafka Protocol Binding        | :x: | :x: |
-| MQTT Protocol Binding         | :x: | :x: |
-| NATS Protocol Binding         | :x: | :x: |
+| Transport Protocols    | [v0.3](https://github.com/cloudevents/spec/tree/v0.3) | [v1.0](https://github.com/cloudevents/spec/tree/v1.0) |
+| ---------------------- | ----------------------------------------------------- | ----------------------------------------------------- |
+| AMQP Protocol Binding  | :x:                                                   | :x:                                                   |
+| HTTP Protocol Binding  | :heavy_check_mark:                                    | :heavy_check_mark:                                    |
+| Kafka Protocol Binding | :x:                                                   | :x:                                                   |
+| MQTT Protocol Binding  | :x:                                                   | :x:                                                   |
+| NATS Protocol Binding  | :x:                                                   | :x:                                                   |
 
 ## Community
 
--- a/src/event/cloudevent.ts
+++ b/src/event/cloudevent.ts
@@ -171,10 +171,11 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
   /**
    * Emit this CloudEvent through the application
    *
-   * @return {CloudEvent} current CloudEvent object
+   * @param {boolean} ensureDelivery fail the promise if one listener fail
+   * @return {Promise<CloudEvent>} this
    */
-  public emit(): this {
-    Emitter.emitEvent(this);
+  public async emit(ensureDelivery = true): Promise<this> {
+    await Emitter.emitEvent(this, ensureDelivery);
     return this;
   }
 
--- a/src/event/cloudevent.ts
+++ b/src/event/cloudevent.ts
@@ -1,4 +1,5 @@
 import { v4 as uuidv4 } from "uuid";
+import { Emitter } from "..";
 
 import {
   CloudEventV03,
@@ -165,6 +166,16 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
         throw new ValidationError("invalid payload", e);
       }
     }
+  }
+
+  /**
+   * Emit this CloudEvent through the application
+   *
+   * @return {CloudEvent} current CloudEvent object
+   */
+  public emit(): this {
+    Emitter.emitEvent(this);
+    return this;
   }
 
   /**
--- a/src/transport/emitter.ts
+++ b/src/transport/emitter.ts
@@ -67,7 +67,7 @@ export class Emitter extends EventEmitter {
   /**
    * Singleton store
    */
-  static singleton: Emitter | undefined = undefined;
+  static instance: Emitter | undefined = undefined;
 
   /**
    * Create an Emitter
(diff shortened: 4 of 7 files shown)


lholmquist:
I think one of the main reasons that we removed the Emitter, was to remove any dependency on  external protocol libraries,  in this case axios.  This also becomes important when adding in new protocols, like MQTT, etc..  We only wanted this module to output the correct formatting and let the user decide which library to use to actually emit it.

I'm not sure how this accomplishes that task

loopingz:
@lholmquist this is not at all pushing to any technologies, it just streamlines the way to send event inside an application, a programmer should not care or know about what its subscribers are to emit an event. Without this kind of logic, it is hard to implement a subscription API, as the subscription is dynamic so is the endpoints to send the event too.

You can see the example on https://gateway.cloudevents.loopingz.com

loopingz:
As @grant comment, I will move from EventEmitter to static async functions

loopingz:
So I added the on() as static to have a shortcut on the Emitter class
I kept the singleton but renamed it to `instance`, as a developer we do not need to get access to the singleton.

I added an option `ensureDelivery` to wait for all listener to process the event if set to true

src/event/cloudevent.ts
```diff
@@ -167,6 +168,17 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
     }
   }
 
+  /**
+   * Emit this CloudEvent through the application
+   *
+   * @param {boolean} ensureDelivery fail the promise if one listener fail
+   * @return {Promise<CloudEvent>} current CloudEvent object
```
lance, line 175: Nit, but more direct?
```suggestion
   * @return {Promise<CloudEvent>} this
```

src/transport/emitter.ts
```diff
@@ -91,16 +92,31 @@ export function emitterFor(fn: TransportFunction, options = { binding: HTTP, mod
...
+  /**
+   * Create an Emitter
+   * On v4.0.0 this class will only remains as Singleton to allow using the
+   * EventEmitter of NodeJS
+   *
+   *
+   * @param {Object} [options] The configuration options for this event. Options
+   * provided will be passed along to Node.js `http.request()`.
+   * https://nodejs.org/api/http.html#http_http_request_options_callback
+   * @deprecated Will be removed in 4.0.0. Consider using the emitterFactory
```
lance, line 116: If the `Emitter` class will remain and extend `EventEmitter`, the constructor can't really be removed, can it? Maybe the text should be something more like this?

 ```suggestion
   * @deprecated in 4.0.0 this class functionality will be limited to events propagated throughout the Node.js process. To emit events over a transport protocol, consider using the emitterFactory
```

File: src/transport/emitter.ts [hunk 3]
```diff
@@ -131,4 +147,48 @@ export class Emitter {
...
+  }
+
+  /**
+   * Add a listener for eventing
+   *
+   * @param {string} event type to listen to
+   * @param {Function} listener to call on event
+   * @return {void}
+   */
+  static on(event: "cloudevent" | "newListener" | "removeListener", listener: (...args: any[]) => void): void {
```
lance, line 170: I'm starting to lean towards allowing `any` in our linter. It makes me jump through ridiculous hoops sometimes to avoid it. In this case, I see you're using the same function signature as the Node.js `EventEmitter` type.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fbf9926cd51d7123e12cac280ccecaac01818bd3/types/node/events.d.ts#L63

We don't need to figure out how to get around this linting problem before this lands. Personally, I'd like to see the rule loosened, but I don't know TS linting rules well enough to loosen it only in certain situations (and I don't know what those would be). Maybe a warning is fine, but we'll be seeing this with every build from now on. :shrug:

src/transport/emitter.ts
```diff
@@ -131,4 +147,48 @@ export class Emitter {
...
+  /**
+   * Emit an event inside this application
+   *
+   * @param {CloudEvent} event to emit
+   * @param {boolean} ensureDelivery fail the promise if one listener fail
+   * @return {void}
+   */
+  static async emitEvent(event: CloudEvent, ensureDelivery = false): Promise<void> {
+    if (!ensureDelivery) {
+      // Ensure delivery is disable so we do wait for Promise
```
lance, line 183: Nit:
```suggestion
      // Ensure delivery is disabled so we don't wait for Promise
```

src/transport/emitter.ts
```diff
@@ -131,4 +147,48 @@ export class Emitter {
...
+      Emitter.getInstance().emit("cloudevent", event);
+    } else {
+      // Execute all listeners and wrap them in a Promise
+      await Promise.all(
+        Emitter.getInstance()
+          .listeners("cloudevent")
+          .map((l) => Promise.resolve(l(event))),
```
lance, line 190: This looks like it will never reject. Shouldn't this be something more like this?
```suggestion
          .map((l) => {
            Promise.resolve(await l(event))),
```
If any of the listeners throw or return a rejected `Promise` that exception will unwrap all the way up to `Promise.all()` which will cause `Promise.reject()`, correct?

lance (changes requested):
Looks good! I really like this idea and think it will work well in certain scenarios - especially the browser. :100:  I have a few minor requested changes, and one question about `Promise` handling in the emitter.

src/transport/emitter.ts
loopingz: It is correct and I added the unit test for it, not sure I understand your suggestion, the Promise.resolve will just make sure if the result is not a Promise to wrap it inside a Promise, and if it sends an exception or reject it is propagated by the Promise.all

src/transport/emitter.ts
loopingz: :+1: I'm onboard with you, I had to create crazy cast on some lines to please the linter, and EventEmitter is defined with any

src/transport/emitter.ts
loopingz: Which reminds me I forgot to add the test :)

src/transport/emitter.ts
loopingz: I added the change to this PR for @typescript-eslint/no-explicit-any

src/transport/emitter.ts
loopingz: I did simplify it to `.map(async (l) => l(event)),`

loopingz: I had to disable the last part of my test as it runs locally fine but not in Github Actions. Its goal is to verify that the promise is not rejected with ensureDelivery=false, but it is normally sent to [EventTarget error handling](https://nodejs.org/api/events.html#events_eventtarget_error_handling)

test/integration/emitter_singleton_test.ts
```diff
@@ -0,0 +1,54 @@
...
+  it("emitWithEnsureDeliveryError", async () => {
+    const emitter = async () => {
+      throw new Error("Not sent");
+    };
+    Emitter.on("cloudevent", emitter);
+    // Should fail with emitWithEnsureDelivery
+    await rejects(() => fixture.emit(true));
+    // Should not fail with emitWithEnsureDelivery
+    // Work locally but not on Github Actions
+    // await doesNotReject(() => fixture.emit(false));
```
lance, line 52: Oh I see now why `doesNotReject` is unused. But I am confused why this fails on GH Actions but not locally (confirmed locally here too).

lance (commented):
LGTM but can you help me understand why this test is failing on GH Actions?

src/transport/emitter.ts
lance: > It is correct and I added the unit test for it, not sure I understand your suggestion

Me neither anymore! LGTM :)

src/event/cloudevent.ts
```diff
@@ -167,6 +168,17 @@ export class CloudEvent implements CloudEventV1, CloudEventV03 {
     }
   }
 
+  /**
+   * Emit this CloudEvent through the application
+   *
+   * @param {boolean} ensureDelivery fail the promise if one listener fail
+   * @return {Promise<CloudEvent>} this
+   */
+  public async emit(ensureDelivery = false): Promise<this> {
```
grant, line 177: Not a fan of `ensureDelivery`. I would expect no parameter. The Promise should fail if there is no delivery and the user should retry the delivery or handle it.

loopingz: The event system of node is not done this way, so if we follow normal events it will never fail, I think adding the option to ensure delivery is useful for some cases

.eslintrc
```diff
@@ -30,6 +30,7 @@
     }],
     "valid-jsdoc": "warn",
     "semi": ["error", "always"],
-    "quotes": ["error", "double", { "allowTemplateLiterals": true }]
+    "quotes": ["error", "double", { "allowTemplateLiterals": true }],
```
grant, line 33: Modifying ESlint rules is great, but really out of scope for a PR about emitters.

I would imagine 2 PRs that are separate:
- One PR the rules for quotes/any.
- One PR that changes the emitter.

README.md
```diff
@@ -1,7 +1,7 @@
 # JavaScript SDK for CloudEvents
 
-[![Codacy Badge](https://api.codacy.com/project/badge/Grade/bd66e7c52002481993cd6d610534b0f7)](https://www.codacy.com/app/fabiojose/sdk-javascript?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=cloudevents/sdk-javascript&amp;utm_campaign=Badge_Grade)
-[![Codacy Badge](https://api.codacy.com/project/badge/Coverage/bd66e7c52002481993cd6d610534b0f7)](https://www.codacy.com/app/fabiojose/sdk-javascript?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=cloudevents/sdk-javascript&amp;utm_campaign=Badge_Coverage)
+[![Codacy Badge](https://api.codacy.com/project/badge/Grade/bd66e7c52002481993cd6d610534b0f7)](https://www.codacy.com/app/fabiojose/sdk-javascript?utm_source=github.com&utm_medium=referral&utm_content=cloudevents/sdk-javascript&utm_campaign=Badge_Grade)
```
grant, line 3: nit: Modifying the URLs and list markdown isn't really in scope for this Emitter PR

test/integration/emitter_factory_test.ts
```diff
@@ -107,7 +107,13 @@ function testEmitter(fn: TransportFunction, bodyAttr: string) {
...
 }
 
-function assertBinary(response: Record<string, string>) {
+/**
+ * Verify the received binary answer compare to the original fixture message
+ *
+ * @param {Record<string, Record<string, string>>} response received to compare to fixture
+ * @return {void} void
+ */
+export function assertBinary(response: Record<string, string>): void {
```
grant, line 116: Sorry, this is not in scope with the PR.

src/transport/emitter.ts
```diff
@@ -58,3 +59,66 @@ export function emitterFor(fn: TransportFunction, options = { binding: HTTP, mod
...
+  }
+
+  /**
+   * Add a listener for eventing
+   *
+   * @param {string} event type to listen to
+   * @param {Function} listener to call on event
+   * @return {void}
+   */
+  static on(event: "cloudevent" | "newListener" | "removeListener", listener: (...args: any[]) => void): void {
```
grant, line 100: Why are we creating an event listening system for an event emitter?

I do not expect an eventing system like this.

test/integration/emitter_singleton_test.ts
```diff
@@ -0,0 +1,56 @@
+import "mocha";
+
+import { emitterFor, HTTP, Mode, Message, Emitter } from "../../src";
+
+import { fixture, assertStructured } from "./emitter_factory_test";
+
+import { rejects, doesNotReject } from "assert";
+
+describe("Emitter Singleton", () => {
```
grant, line 9: The casts used in this test seem very fishy (casting to unknown, then an interface).

Additionally, there are no comments, making the tests hard to understand.

test/integration/emitter_singleton_test.ts
```diff
@@ -0,0 +1,56 @@
...
+  it("emit", async () => {
+    let msg: Message | unknown = undefined;
+    const fn = async (message: Message) => {
+      msg = message;
+    };
+    const emitter = emitterFor(fn, { binding: HTTP, mode: Mode.STRUCTURED });
+    Emitter.on("cloudevent", emitter);
+    fixture.emit();
+    while (msg === undefined) {
```
grant, line 18: A while loop like this, waiting for a response, is really bad practice. I'd imagine an await pattern after emitting.

grant (changes requested):
Sorry, I don't believe this change is beneficial. There are numerous issues I've pointed out in comments in a review, including changes not directly related to emitter functionality.

Main issue is that we removed a static emitter (mentioned here https://github.com/cloudevents/sdk-javascript/pull/366#issuecomment-726180477) but we have a `emitterFor` function for this use-case.

test/integration/emitter_singleton_test.ts
loopingz: I agree overall but...
The NodeJS eventing system is not designed with Promise https://nodejs.org/api/events.html

test/integration/emitter_singleton_test.ts
loopingz: That is why we removed the eslint rule according with @lance

test/integration/emitter_factory_test.ts
loopingz: The export was added to be able to share some logics and avoid duplicated code

.eslintrc
lance: > I would imagine 2 PRs that are separate:

I think the point of this change is that the existing linter rules was causing this PR to generate warnings during the build. See https://github.com/cloudevents/sdk-javascript/pull/366#discussion_r522475661. I'm not sure of the benefit returned by extracting this small change into a separate PR, landing that first, then landing this. In fact, I recall recently that you were not happy about the number of commits in this repo - this seems counter to that complaint! :wink:

README.md
lance: I tend to agree with this, but don't find it objectionable.

test/integration/emitter_factory_test.ts
lance: > The export was added to be able to share some logics and avoid duplicated code

Makes sense to me.

lance (approved):

.eslintrc
grant: Hey, I'd quickly approve a change/fix to the eslint rules, but I don't want to approve a change that includes `while (msg === undefined) {`. The eslint change is a different decision than emitter functionality and would like to be merged even if any emitter PRs were not merge. Does that make sense?

README.md
loopingz: The markdown linter should be also set to avoid this in the future

src/transport/emitter.ts
loopingz: This is a shortcut to avoid the `getSingleton` that you did not like, I started to reimplement the EventEmitter interface, but found it overkill, as there are several other methods that can be useful. So instead of reinventing the wheel, I just added a shortcut method to streamline the process

.eslintrc
lance: > The eslint change is a different decision than emitter functionality and would like to be merged even if any emitter PRs were not merge. Does that make sense?

Sure. It just seems pedantic to demand this.

test/integration/emitter_singleton_test.ts
grant: I'm sorry, but there are better methods to test an asynchronous pattern than a while undefined loop. A while loop like this is blocking I/O.

If we did want to go this route, in Node, it's common to not write `=== false` or `=== undefined`, rather use  `!`. If we can, I'd `await` the event.

loopingz: You mean this one: https://nodejs.org/api/events.html#events_events_on_emitter_eventname_options because it is not available on Node 10 which is one of the target

I do not know other way to get a promise on an event, how would you do it?

src/transport/emitter.ts
lance: In my opinion, [the example added to the README](https://github.com/cloudevents/sdk-javascript/pull/366/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R96-R114) is very illustrative. This PR makes it so that you can set up your emitter function using `emitterFor()`, then have that respond to all `emit` events in the Node event system by actually sending the `CloudEvent`. To me that seems quite valuable. Imagine that you have a large system that is pushing out `CloudEvents`, perhaps even to multiple different endpoints. You could add multiple emitter functions at startup that each send to a different endpoint. Then by simply calling `new CloudEvent({...}).emit();` the `CloudEvents` is sent - with that single call - to all endpoints. I like it a lot.

test/integration/emitter_singleton_test.ts
loopingz: Also, can you iterate on the blocking I/O as this loop include the await of timeout so I wonder why/how it is blocking the I/O.
Thanks for clarifying so I can get a better understanding of NodeJS

.eslintrc
grant: Lance, I'm not demanding. This is just a review suggestion.

test/integration/emitter_singleton_test.ts
grant: Right now the `on` function implemented does not return a `Promise`. Promises are a good way to have async processing in general.

You could use the Events emitter.on pattern too. Whatever the choice, I would not have a while loop like this, or would imagine exponential backoff if required.

src/event/cloudevent.ts
lance, line 177: I wonder if this shouldn't default to `true`. Doesn't it seem that it would typically be bad practice to "fire and forget"?
```suggestion
  public async emit(ensureDelivery = true): Promise<this> {
```

src/transport/emitter.ts
loopingz: Yes this was the exact purpose, thanks for summarizing it so well

test/integration/emitter_singleton_test.ts
```diff
@@ -0,0 +1,56 @@
+import "mocha";
+
+import { emitterFor, HTTP, Mode, Message, Emitter } from "../../src";
+
+import { fixture, assertStructured } from "./emitter_factory_test";
+
+import { rejects, doesNotReject } from "assert";
+
+describe("Emitter Singleton", () => {
+  it("emit", async () => {
```
lance, line 10: Maybe some more descriptive text for the test case would be good.

```suggestion
  it("emits a Node.js 'cloudevent' event as an EventEmitter", async () => {
```

test/integration/emitter_singleton_test.ts
loopingz: The NodeJS pattern is available only on Node v12.16.0 so cannot be used.
I just updated the PR with a custom Promise that will resolve by the event itself mimicking a bit the pattern

src/transport/emitter.ts
grant: Yup, the README example is great.

```js
const axios = require("axios").default;
const { emitterFor, Mode, CloudEvent, Emitter } = require("cloudevents");
function sendWithAxios(message) {
  // Do what you need with the message headers
  // and body in this function, then send the
  // event
  axios({
    method: "post",
    url: "...",
    data: message.body,
    headers: message.headers,
  });
}
const emit = emitterFor(sendWithAxios, { mode: Mode.BINARY });
// Set the emit
Emitter.on("cloudevent", emit);
...
// In any part of the code will send the event
new CloudEvent({ type, source, data }).emit();
```

I think the fundamental question is why are we adding back the `emit` function? And adding a `ensureDelivery` param?

A CloudEvent object and a CloudEvent emitter/receiver seem like separate concepts.

```js
const axios = require("axios").default;
const { emitterFor, Mode, CloudEvent, Emitter } = require("cloudevents");
function sendWithAxios(message) {
  // Do what you need with the message headers
  // and body in this function, then send the
  // event
  axios({
    method: "post",
    url: "...",
    data: message.body,
    headers: message.headers,
  });
}
const emit = emitterFor(sendWithAxios, { mode: Mode.BINARY });

// In any part of the code will send the event
const ce = new CloudEvent({ type, source, data });
emit(ce);
```

Folks can add an eventing layer on-top of the primitives here. Maybe point out an error in the above alternative code?

.eslintrc
loopingz: The goal was to merge the PR in a timely manner so at that time we thought it would make sense to have only one PR

src/transport/emitter.ts
loopingz: The emitEvent function was adding because EventEmitter is not able to "wait" for all listeners to finish async jobs.
You might by design consider that you need to fail the whole process if you were not able to send the event, or you can decide to let it go depending on how critical to your business the CloudEvent is.
This is the SDK so I think it should be able to answer both needs instead of limiting to one interpretation.

test/integration/emitter_singleton_test.ts
loopingz: It is part of the wider group `EventEmitter`

src/event/cloudevent.ts
loopingz: I put false by default as this is the behavior of EventEmitter, I suppose both are defendable, if you have several subscribers maybe only is critical.
I suppose in my future implementation I would go for `true` with the main subscriber and optional one

.eslintrc
grant: Having a PR reviewed in a timely manner is fair. In general, it would be easier to not go back and forth and review faster with a GitHub issue and small PR. There have been discussions about removing this before, and we just talked about splitting this repo into different modules, so it's difficult to approve this PR right away.

That said, this repo is likely going to change a bit with the split, and I don't want to block you if Lance approves.

grant:
This repo is likely going to change a bit with the split we talked about, and I don't want to block emitting efforts. Will dismiss my review.

That said, I think more discussion about event listeners / lifecycle methods in general in the CE SDKs would probably be worthwhile next call. I hope the points I made are clear.

```ts
static on(event: "cloudevent" | "newListener" | "removeListener", listener: (...args: any[])
```

src/transport/emitter.ts
lance: > So I wouldn't imagine a CloudEvent has a send/receive function.

Just to be clear, we are talking about two separate concepts. It's easy to get things mixed up when all of the terms are so similar. The two separate things as I see it are:

1. Sending a `CloudEvent` over some network transport using the `emitterFor()` factory to create a user defined function that is capable of converting a `CloudEvent` into a `Message` and sending that over the wire with a network transport (typically HTTP).
1. Using the Node.js built in `EventEmitter` capabilities to notify a [`listener`](https://nodejs.org/api/events.html#events_emitter_on_eventname_listener) that a `CloudEvent` should be sent using the method described in 1) point above. The `listener` in this case is the transport function created with `emitterFor()`. The new `CloudEvent#emit()` function just facilitates sending these using the `Emitter.emitEvent()` function.

This pull request is about using what is described in 2) to facilitate and orchestrate 1) in an efficient way across a large system that may need to send `CloudEvents` over network transport to multiple endpoints and do so efficiently by making use of built in Node.js `EventEmitter` capabilities.

> Maybe point out an error in the above alternative code?

This pull request is not an alternative to that code. It builds upon that code to add a very useful feature, in my opinion, with very low overhead using Node.js built in, idiomatic capabilities.

loopingz:
I think we are all set, thanks gentlemen

@grant, sure let's discuss this longer in our call, we can still fix or improve things afterward

I, on purpose, split into two commits to avoid having the linter modification blended in the feature, so please do a not splash commit on merge

Issue #385, opened 2021-02-17
Class static side 'typeof Emitter' incorrectly extends base class static side 'typeof EventEmitter'


moltar:
**Describe the Bug**

When building the project with TS, I get the following error:

```
node_modules/cloudevents/dist/transport/emitter.d.ts:48:22 - error TS2417: Class static side 'typeof Emitter' incorrectly extends base class static side 'typeof EventEmitter'.
  Types of property 'on' are incompatible.
    Type '(event: "newListener" | "removeListener" | "cloudevent", listener: (...args: any[]) => void) => void' is not assignable to type '(emitter: EventEmitter, event: string) => AsyncIterableIterator<any>'.
      Types of parameters 'event' and 'emitter' are incompatible.
        Type 'EventEmitter' is not assignable to type '"newListener" | "removeListener" | "cloudevent"'.
          Type 'EventEmitter' is not assignable to type '"cloudevent"'.

48 export declare class Emitter extends EventEmitter {
                       ~~~~~~~
```

```
    "cloudevents": "4.0.0",
    "@types/node": "14.14.27",
    "typescript": "4.1.3"
```

**Steps to Reproduce**

N/A

**Expected Behavior**

Build passing.

**Additional context**

```
    "cloudevents": "4.0.0",
    "@types/node": "14.14.27",
    "typescript": "4.1.3"
```

lance:
Hi @moltar if you could provide some steps to reproduce this, that would be great! Looking at the typescript example in this repository, I do not see this problem. Thanks. :)

moltar:
Yup, here you go:

https://github.com/moltar/cloudevents-exp

Thanks!

hilfor:
+1

mnesbitt:
I'm having this issue as well in version 4.0.0.

I did some testing and it looks like it's caused by the version of `@types/node` being used. The typescript example in this repository is using `@types/node@8.9.0` which is 3 years old. I'm currently using `@types/node@14.14.10` and seeing the issue, and @moltar 's example is using 14.14 as well.

I downgraded to `cloudevents@3.2.0` for now and it's compiling for me with the latest `@types/node`.

lance:
@mnesbitt thanks for that info. I'll take a look.

lance:
@mnesbitt @moltar @hilfor if you all have an opportunity to test [this change](https://github.com/cloudevents/sdk-javascript/pull/387) in your project and provide feedback, that would be appreciated. Thanks!

[...]

PR #387, opened 2021-02-25
fix: Emitter should not extend EventEmitter

lance:
This change modifies Emitter so that it does not directly extend the Node.js
EventEmitter class. Instead, it holds a singleton instance of an EventEmitter
but is not an instance of EventEmitter itself.

This commit also updates the typescript example to use a modern version of
@types/node and typescript.

Finally there are a few minor formatting changes picked up by eslint.

@loopingz if you have a chance to review this, that would be great. Thanks!

Fixes: https://github.com/cloudevents/sdk-javascript/issues/385

### DIFF
--- a/src/transport/emitter.ts
+++ b/src/transport/emitter.ts
@@ -88,8 +88,8 @@ export class Emitter {
    * @param {Function} listener to call on event
    * @return {void}
    */
-  static on(event: string, listener: (...args: any[]) => void): void {
-    this.getInstance().on(event, listener);
+  static on(event: "cloudevent" | "newListener" | "removeListener", listener: (...args: any[]) => void): void {
+    Emitter.getInstance().on(event, listener);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


grant (approved):

src/transport/emitter.ts
```diff
@@ -97,7 +88,7 @@ export class Emitter extends EventEmitter {
    * @param {Function} listener to call on event
    * @return {void}
    */
-  static on(event: "cloudevent" | "newListener" | "removeListener", listener: (...args: any[]) => void): void {
```
loopingz, line 100: I liked the idea of typing more precisely than the generic string, I checked to revert this on your branch, and seems ok

src/transport/emitter.ts
```diff
@@ -97,7 +88,7 @@ export class Emitter extends EventEmitter {
...
+  static on(event: string, listener: (...args: any[]) => void): void {
     this.getInstance().on(event, listener);
```
loopingz, line 92: While rereading the source code, I think it would be more consistent to use `Emitter.getInstance()` on this line too

loopingz (commented):
A part from the 2 small comments, it looks good to me.

Thanks

lholmquist (approved):

loopingz (approved):

loopingz (approved):

PR #394, opened 2021-03-24
fix: ensure loose validation for isEvent and toEvent

lance:
The `HTTP.isEvent()` and `HTTP.toEvent()` functions both had some validation
code that violated the principle of loose validation when receiving an
event over HTTP. As discussed, the validation should be managed by the
receiver in this case with `event.validate()`

### DIFF
--- a/src/message/http/index.ts
+++ b/src/message/http/index.ts
@@ -1,7 +1,14 @@
 import { CloudEvent, CloudEventV03, CloudEventV1, CONSTANTS, Mode, Version } from "../..";
 import { Message, Headers } from "..";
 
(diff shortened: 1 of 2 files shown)


lance:
@lholmquist I was hoping to get this in before 4.0.1 but :shrug:. Once this is out, I can replace all of this with `HTTP.isEvent()`. https://github.com/boson-project/faas-js-runtime/blob/e06ee4297e6a806ca1df0ac2b075bae7cdd289f0/lib/event-handler.js#L34-L50

lholmquist (approved):

PR #405, opened 2021-04-20
chore: fix axios dependency

lance:
Somehow axios had been completely removed from the dependencies in package.json. I guess our tests still worked because @types/axios must have a dependency on axios, and by the transitive power of npm dependencies it was there. But we actually have axios in the package itself. So, users who want to use the axios emitter that we ship will have to know to also install axios without this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1470,10 +1470,9 @@
       "dev": true
     },
     "axios": {
-      "version": "0.21.0",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.0.tgz",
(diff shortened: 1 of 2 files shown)


lholmquist:
I didn't realize we still had an axios emitter.  I thought we were trying to move away from including any specific protocol libraries.

lance:
> I didn't realize we still had an axios emitter. 

https://github.com/cloudevents/sdk-javascript/blob/cc43f3bd10caffbc47b9ce42df71782177176001/src/transport/http/index.ts#L1-L17

> I thought we were trying to move away from including any specific protocol libraries.

Yes we are. But we forgot to actually remove it for 4.0.

> The BEFORE functionality still exists until the release of 4.x at which time it will be removed. 

https://github.com/cloudevents/sdk-javascript/pull/342

I mean, I guess we could leave it as-is and consider it "removed" - though it is still accessible, and could work fine as long as the user declares the axios dependency themselves.

lance:
Closing this for now. We can revisit later if we need to.

PR #427, opened 2021-08-04
chore: use git submodules for conformance tests

lance:
I don't think downloading to `/tmp` for each `npm test` is such a great
idea. This does mean that contributors to this repo will need to run the
following command once on their clone after this commit lands.

```
git submodule init
git submodule update
```


lholmquist (approved):
I'm not usually a fan of submodules, but this case makes sense.  LGTM

PR #429, opened 2021-08-26
fix:  update express example with framework features.

### LINKED ISSUE #379: Use HTTP.structured() and/or HTTP.binary() in Express.js example
**The Express.js example is only partially correct**

In the Express.js example, we are doing some things by hand that should really be handled by the framework. 

https://github.com/cloudevents/sdk-javascript/blob/6be3b2751401e529fce6ccf8a406a057472c10a9/examples/express-ex/index.js#L21-L41

In the response, we are just returning the `CloudEvent` instead of using the builtin framework capabilities. For example.

```
 const responseEventMessage = new CloudEvent({
        ...receivedEvent,
        source: '/',
        type: 'event:response'
    });
    // const message = HTTP.binary(responseEventMessage)
    const message = HTTP.structured(responseEventMessage)
    res.set(message.headers)
    res.send(message.body)
```

**Expected Behavior**

The example should more accurately represent the capabilities of the framework.

**Additional context**

See: https://stackoverflow.com/questions/65854143/how-to-write-knative-sequence-step

lholmquist:
fixes #379

### DIFF
--- a/examples/express-ex/index.js
+++ b/examples/express-ex/index.js
@@ -33,7 +33,12 @@ app.post("/", (req, res) => {
     responseEventMessage.data = {
       hello: 'world'
     };
-    res.status(201).json(responseEventMessage);
+
+    // const message = HTTP.binary(responseEventMessage)
+    const message = HTTP.structured(responseEventMessage)
+    res.set(message.headers)
+    res.send(message.body)
+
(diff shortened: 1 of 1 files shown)


lance (approved):
/lgtm

PR #435, opened 2021-09-13
Remove node 10 from ci

lholmquist:
This removes Node 10 from the CI since it is no longer supported.

This commit i used will only create a patch release,  but we might consider a major version bump for getting rid of a node version


lance:
I think we are already at a major version bump anyway. https://github.com/cloudevents/sdk-javascript/pull/430

PR #443, opened 2021-11-19
chore: remove uuid dependency in favor of crypto package

lance:
## Proposed Changes

As of Node.js 14.17.0 the crypto package has a `randomUUID()` function
which we can use to replace our external dependency.

### DIFF
--- a/.github/workflows/nodejs-ci-action.yml
+++ b/.github/workflows/nodejs-ci-action.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: ["14.x", "16.x"]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


lance:
Note that this change removes support for Node.js 12.x which is not yet EOL. We may want to consider holding off on this until April 2022 when that happens.

matthewrobertson:
I would probably wait a while after Node.js 12 goes EOL before you land this so developer have time to migrate.

grant (changes requested):
As mentioned, I don't recommend removing support for Maintenance LTS versions of Node. This is a good reference, but we shouldn't remove this version from support or tests.

For example, many customers are still running Node 10 applications and haven't upgraded their Node version.

lance:
> As mentioned, I don't recommend removing support for Maintenance LTS versions of Node. This is a good reference, but we shouldn't remove this version from support or tests.
> 
> For example, many customers are still running Node 10 applications and haven't upgraded their Node version.

Yikes! That's been EOL for a while. In any case, y'all are probably right. I was just eager to show that we don't like to carry a lot of dependencies and a willingness to remove them. :)

Issue #441, opened 2021-11-17
CloudEvent with no datacontenttype parsed as JSON instead of plain text


csantanapr:
**Describe the Bug**
Doing a demo based on this blog to consume S3 events from Kamelet
The nodejs SDK can't handle the HTTP POST request with cloud event
Using the following code https://github.com/knative/docs/blob/main/docs/serving/samples/cloudevents/cloudevents-nodejs/index.js

**Steps to Reproduce**
1. create nodejs service
2. create trigger to send cloudevent from broker to nodejs service

**Expected Behavior**
That the cloud event is properly handle.
The knative event-display using the goland cloudevent can handle plaintext in body and missing content-type in header.
https://github.com/knative/eventing/blob/main/cmd/event_display/main.go#L59

**Additional context**

Try both versions of the SDK 4.0.3 and 5.0.0 same error

The input event from Camel Kamelet S3 is the following
```
Debug Incoming Request: {
   "headers": {
     "host": "func1.default.svc.cluster.local",
     "user-agent": "Go-http-client/1.1",
     "content-length": "11",
     "accept-encoding": "gzip",
     "ce-id": "A8A6BD24B2AE5FF-0000000000000017",
     "ce-knativearrivaltime": "2021-11-17T22:48:50.315578709Z",
     "ce-source": "source",
     "ce-specversion": "1.0",
     "ce-time": "2021-11-17T22:48:50.306Z",
     "ce-type": "org.apache.camel.event",
     "forwarded": "for=10.131.0.156;proto=http",
     "k-proxy-request": "activator",
     "traceparent": "00-0f3421c775056be4f762f87b5cee145d-5f24409ef4edc327-00",
     "x-forwarded-for": "10.131.0.156, 10.128.2.48",
     "x-forwarded-proto": "http",
     "x-request-id": "13396d30-5487-46c1-9d21-b3fd675447e0"
   },
   "body": "test func1\n"
 }
```
The error is the following
```
SyntaxError: Unexpected token 
  in JSON at position 11
     at JSON.parse (<anonymous>)
     at parseJSON (/opt/app-root/src/node_modules/cloudevents/dist/parsers.js:39:67)
     at JSONParser.parse (/opt/app-root/src/node_modules/cloudevents/dist/parsers.js:40:16)
     at parseBinary (/opt/app-root/src/node_modules/cloudevents/dist/message/http/index.js:166:23)
[...]

PR #444, opened 2021-11-19
fix: do not assume an empty content-type header is JSON

lance:
The parser for HTTP binary made the assumption that if there was no `content-type`
header in the incoming message, it should inject `application/json`. Discussion
about the rationale for this is in https://github.com/cloudevents/sdk-javascript/issues/441.

This commit removes that injection and adds a test to ensure the bytes are not parsed, but just passed along untouched.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/441

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,14594 @@
 {
   "name": "cloudevents",
   "version": "5.0.0",
-  "lockfileVersion": 1,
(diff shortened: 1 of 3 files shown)


lance:
@grant or @lholmquist would one of you ptal :smiley_cat:

lholmquist (approved):

PR #453, opened 2021-12-15
chore: update cucumber dependency and remove prettier

lance:
The combination of prettier and eslint was causing some conflicting error
messages in formatting between VSCode and using npm in the CLI. For the most
part, there were only a couple of required formatting changes that prettier
was covering, so the change is minor.

The cucumber dependency had a major version bump and was carrying some unsafe
dependencies in the older version. This commit bumps to the new version and
makes appropriate configuration changes.


lance:
@grant @lholmquist can either of you take a quick look here? Thanks!

grant (approved):

Issue #390, opened 2021-03-06
plans for amqp/nats/stan/etc (similar to sdk-go)?


justinrush:
**Is your feature request related to a problem? Please describe.**
The sdk-go project contains protocol implementations for other popular message buses, such as AMQP, NATS/STAN, and Kafka. I would like to see the same protocol layer available in the javascript SDK.

**Describe the solution you would like to see**
The same interfaces created for Protocol, Consumer, Receiver, Sender, etc, with a client class that leverages the interfaces. Bonus points if someone can write one or more of the protocol implementations for AMQP or Kafka.

**Additional context**
I'm working on a project that has a mix of typescript/javascript and golang based services using a mix of message transports. It would be great to standardize on something like cloudevents for the message format.

lance:
Hi @justinrush - yes, we would definitely like to support all of the protocols that have a [specification](https://github.com/cloudevents/spec/#cloudevents-documents). There is no active work toward these goals at the moment, but we welcome contributions if you find yourself interested in that.

Alexis01:
what about this request?

LeoVS09:
@lance @justinrush @Alexis01 My team wants to use CoudEvents for our JS services, but currently, for transport, we use only Kafka. I made [cloudevents-kafka](https://github.com/LeoVS09/cloudevents-js-sdk-kafka) package as Kafka transport extension for the main `cloudevents` library, with all features which need for us right now. 
I do not have enough time to work on PR in this repository, but anyone who wants can use my package or copy it as PR in this repo.

lholmquist:
once we do a new release with. #455 ,  can this be closed, or should we keep it open until those other protocols are dealt with?

lance:
I think we can probably just close this, period. It's pretty broad and includes more than just transport implementations. The original ask is for all of the same interfaces here as in the Go SDK. But this implementation comes at it from a completely different perspective - not the least of which is the fact that Javascript is 'bring-your-own-transport-implementation' and Go just uses the built in HTTP under the hood to provide the implementation directly.

lholmquist:
closing,  and if we need to re-open we can

PR #455, opened 2021-12-22
feat: add support for kafka transport

lance:
This commit extends the `message` package to include Kafka transport. The tests included here were almost entirely copied verbatim from `message_tests.ts` which tests the `HTTP` message binding. The primary change to these tests was the use of the `Kafka` binding instead of the `HTTP` binding. It would be nice, I think, to have generic tests that can test any new transport layer by just taking the binding as a parameter. But that's not what I've done here. :) I've mostly just copied and pasted.

There is one test that is pending as it's not passing yet, and I figured it was a bit of an edge case and could be addressed later.

```
      - Converts base64 encoded data to binary when deserializing structured messages
```

Note that only JSON encoding is supported.

Additionally, some of the type information has changed across the project to more accurately reflect the type of `Message` with `Message<T>`.

Related: https://github.com/cloudevents/sdk-javascript/issues/390


src/message/index.ts
```diff
@@ -41,9 +41,9 @@ export interface Headers extends IncomingHttpHeaders {
  * @property {@linkcode Headers} `headers` - the headers for the event Message
  * @property string `body` - the body of the event Message
```
grant, line 42: Looks like these docs should be updated with the code change.

Maybe other docs too.

src/message/index.ts
```diff
@@ -41,9 +41,9 @@ export interface Headers extends IncomingHttpHeaders {
...
  */
-export interface Message {
+export interface Message<T> {
```
grant, line 44: I believe all generic types require a default value to not be a breaking changes.

```ts
export interface Message<T = string> {
    headers: object;
    body: T | Buffer | unknown;
}
```

If this is a breaking change, please change the conventional commit type to `feat!`.

src/message/kafka/headers.ts
```diff
@@ -0,0 +1,60 @@
...
+ Copyright 2021 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+import { CloudEvent, CONSTANTS, Headers } from "../..";
+
+/**
+ * The set of CloudEvent headers that may exist on a Kafka message
+ */
+export const KAFKA_CE_HEADERS: { [key: string]: string } = Object.freeze({
```
grant, line 11: I believe the type could be stronger or be omitted, specifically the `key: string` type.

```ts
Readonly<{
    ID: string;
    TYPE: string;
    SOURCE: string;
    SPEC_VERSION: string;
    TIME: string;
    SUBJECT: string;
    DATACONTENTTYPE: string;
    DATASCHEMA: string;
}>
```

Otherwise, I'm not getting autocompletion for `KAFKA_CE_HEADERS.`...

src/message/kafka/index.ts
```diff
@@ -0,0 +1,248 @@
...
+ * @param {CloudEvent<T>} event The event to serialize
+ * @returns {KafkaMessage<T>} a KafkaMessage instance
+ */
+function toBinaryKafkaMessage<T>(event: CloudEvent<T>): KafkaMessage<T> {
+  // 3.2.1. Content Type
+  // For the binary mode, the header content-type property MUST be mapped directly
+  // to the CloudEvents datacontenttype attribute.
+  const headers: Headers = { 
+    ...{ [CONSTANTS.HEADER_CONTENT_TYPE]: event.datacontenttype }, 
+    ...headersFor(event) };
```
grant, line 52: nit: I'm surprised linting does not move `};` to a new line.

src/message/kafka/index.ts
```diff
@@ -0,0 +1,248 @@
...
+    throw new ValidationError("Value is null or undefined");
+  }
+  if (!message.headers) {
+    throw new ValidationError("Headers are null or undefined");
+  }
+  const cleanHeaders: Headers = sanitize(message.headers);
+  const mode: Mode = getMode(cleanHeaders);
+  switch (mode) {
+    case Mode.BINARY:
+      return parseBinary(message as KafkaMessage<T>);
```
grant, line 108: Type assertion `message as KafkaMessage<T>` can be done once (or fewer times) above.

src/message/kafka/index.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+/**
+ * Creates a Kafka key for the event provided
+ * @param {CloudEvent} e an event
+ * @returns {string} the Kafka key
+ */
+ function createKey<T>(e: CloudEvent<T>): string {
+  return `${e.source}/${e.id}`;
+}
```
grant, line 139: nit: indentation looks off in this section

test/integration/emitter_factory_test.ts
```diff
@@ -37,11 +37,11 @@ export const fixture = new CloudEvent({
   [ext3Name]: ext3Value,
 });
 
-function axiosEmitter(message: Message, options?: Options): Promise<unknown> {
```
grant, line 40: It should be a goal to not require parameterization for this file. Can we try adding this feature without a change to this file?

test/integration/emitter_singleton_test.ts
```diff
@@ -12,36 +12,36 @@ import { fixture, assertStructured } from "./emitter_factory_test";
 import { rejects, doesNotReject } from "assert";
 
 describe("Emitter Singleton", () => {
```
grant, line 14: It should be a goal to not require parameterization for this file. Can we try adding this feature without a change to this file?

A separate test for parameterized messages would be fine if needed.

grant (changes requested):
Few requested changes. Hope the comments are helpful.

Nothing particularly blocking.

I'm not familiar with the goal of the SDKs in terms of supporting all protocols / transports. It be best to keep other transport logic to child helpers since a developer will usually only use the SDK for one protocol at a time.

lance:
Thanks for the feedback @grant - I've made some updates. Making the generic type `T` default to `string` was quite helpful - thanks for pointing that out.

lance:
> I'm not familiar with the goal of the SDKs in terms of supporting all protocols / transports. It be best to keep other transport logic to child helpers since a developer will usually only use the SDK for one protocol at a time.

Not sure how I feel about this. Other SDKs do support multiple protocols, and all do so within a single repository. Distributing the bits of these implementations is, of course, language specific. So, in Java, for example, you can pull in just the Kafka bindings iiuc, but that's not the case in Rust, for example, where the implementation  is all together. In any case, the different transport protocol support is all implemented in a single github repo - which is what we would need to do as well. I think the additional bytes that are added to the package by these two files are really not an issue compared to the complexity of reorganizing this repo to be a mono-repo, supporting multiple npm modules. My 2c.

src/message/kafka/index.ts
lance: I tried to simplify this, but didn't have much luck at an elegant solution. If you can suggest a specific change for this that compiles I'd be happy to use it.

grant:
> > I'm not familiar with the goal of the SDKs in terms of supporting all protocols / transports. It be best to keep other transport logic to child helpers since a developer will usually only use the SDK for one protocol at a time.
> 
> Not sure how I feel about this. Other SDKs do support multiple protocols, and all do so within a single repository. Distributing the bits of these implementations is, of course, language specific. So, in Java, for example, you can pull in just the Kafka bindings iiuc, but that's not the case in Rust, for example, where the implementation is all together. In any case, the different transport protocol support is all implemented in a single github repo - which is what we would need to do as well. I think the additional bytes that are added to the package by these two files are really not an issue compared to the complexity of reorganizing this repo to be a mono-repo, supporting multiple npm modules. My 2c.

Yeah it's fine imo. The main thing is being wary of extra dependencies that may be unnecessary, and ensuring it's still easy to contribute to HTTP features without having to affect other protocols.

src/message/kafka/index.ts
```diff
@@ -0,0 +1,249 @@
...
+    headers[CONSTANTS.HEADER_CONTENT_TYPE]?.startsWith(CONSTANTS.MIME_CE_BATCH) as boolean; // A batch of events
+}
+
+/**
+ * Creates a Kafka key for the event provided
+ * @param {CloudEvent} e an event
+ * @returns {string} the Kafka key
+ */
+function createKey<T>(e: CloudEvent<T>): string {
+  return `${e.source}/${e.id}`;
```
LeoVS09, line 139: I am glad you used my implementation, but I initially incorrectly implemented the specification.
`Key` cannot be constructed that way. Correct implementation require to use [Partition Key Extenstion](https://github.com/cloudevents/spec/blob/v1.0.1/kafka-protocol-binding.md#31-key-mapping).
I added that extension in [this commit](https://github.com/LeoVS09/cloudevents-js-sdk-kafka/commit/aed9bb89b21b59fa023ef94d94f6054d6004c622)

**It is important to not construct** `key` in such a way not only because of specification but also because this key will be used for partitioning by Kafka. It must be **not unique**. Different events, the order of which is important, must share the same `key`, and still have a unique `id` for correctly processing the Exactly-Once Delivery flow. 
**In simple words:** better to keep this field empty than try to construct it.

LeoVS09 (changes requested):
Sorry for breaking into your working process, but it critical bug was my failure, and it can create complex problems for production applications

src/message/kafka/index.ts
lance: @LeoVS09 thank you very much for the feedback. I will make the necessary changes for this in the new year. Your implementation has been a very helpful inspiration for this addition and I appreciate both the time you've committed to the code, and that which you spent reviewing the pr. Thanks!

lance:
@LeoVS09 I've made the appropriate changes, ptal.
@grant @lholmquist a bit has changed sine your last reviews, would you mind taking another look at this? Thanks.

lance:
Marking as WIP until https://github.com/cloudevents/sdk-javascript/pull/457 is resolved as this PR depends on many of those changes, and I think it's better to keep the PRs small and focused vs. cramming all of that in here.

lance:
I have rebased with the latest changes from `main` and generally cleaned this up a bit. It's ready for a final review @grant and/or @lholmquist. Thanks!

grant (approved):

PR #457, opened 2022-01-04
chore(refactor): prefer interfaces over concrete classes

lance:
This change modifies the protocol binding interfaces such as `Binding`, `Serializer` and the `cloneWith()` methods to use the `CloudEventV1` interface instead of the implementation class `CloudEvent`. This should make extending the interfaces simpler as using interfaces decouples Specification Extensions for relying on the concrete `CloudEvent` class. This work has grown out of efforts around the implementation of a second transport interface, Kafka.

See: https://github.com/cloudevents/sdk-javascript/pull/455/

In addition, this PR includes:

- The addition of a generic type to the `Message` interface, defaulting to `string`
- Some minor clean up involving what is exported from the `message/http` modules. Now, instead of exporting the entire implementation, only the `HTTP` binding implementation is exported
- A static `CloudEvent.cloneWith()` method has been added which the instance methods now use

I'm pretty sure all of this is fully backwards compatible as there were no changes required for the tests.

I will hold on #455 until this has landed, so that I can pull in these changes and keep 455 small-ish.

### DIFF
--- a/src/event/cloudevent.ts
+++ b/src/event/cloudevent.ts
@@ -207,7 +207,7 @@ See: https://github.com/cloudevents/spec/blob/v1.0/spec.md#type-system`);
    * @return {CloudEvent} returns a new CloudEvent
    */
   public cloneWith<D>(options: Partial<CloudEventV1<D>>, strict = true): CloudEvent<D | T> {
-    return CloudEvent.cloneWith(this as CloudEvent<T>, options, strict);
+    return CloudEvent.cloneWith(this, options, strict);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/event/cloudevent.ts
```diff
@@ -147,7 +147,7 @@ See: https://github.com/cloudevents/spec/blob/v1.0/spec.md#type-system`);
   toJSON(): Record<string, unknown> {
     const event = { ...this };
     event.time = new Date(this.time as string).toISOString();
-    event.data = !isBinary(this.data) ? this.data : undefined;
+    event.data = this.#_data;
```
lance, line 150: Note: It's not clear why we originally were unsetting the `data` property in this function if it was in binary form. But I don't think we should be doing that. Making this change had no effect on the tests. Happy to discuss what the reason is for this code as it was if anyone recalls.

lholmquist (approved):

grant (approved):

PR #459, opened 2022-01-11
feat: add MQTT transport messaging

lance:
## Proposed Changes

Add MQTT as a `Message` format.

## Description

This commit adds `MQTT` to the supported transport protocols by adding a `Binding` and the `MQTTMessage<T>` type, extending the base `Message` type, adding the MQTT fields for `payload`, `PUBLISH` and `User Properties`. The `payload` field directly maps to `Message#body`, while `User Properties` roughly maps to `Message#headers`, even though the properties here are not formatted with a `ce-` prefix like other transport protocols. This is per the spec. See: https://github.com/cloudevents/spec/blob/v1.0.1/mqtt-protocol-binding.md. The `PUBLISH` property is there primarily to support the `"Content type"` and perhaps should be typed better. Feedback welcome.

### DIFF
--- a/src/message/mqtt/index.ts
+++ b/src/message/mqtt/index.ts
@@ -8,14 +8,35 @@ import { Binding, Deserializer, CloudEvent, CloudEventV1, CONSTANTS, Message, Va
 export {
   MQTT,
   MQTTMessage,
+  MQTTMessageFactory
 };
 
+/**
+ * Extends the base {@linkcode Message} interface to include MQTT attributes, some of which
+ * are aliases of the {Message} attributes.
+ */
 interface MQTTMessage<T> extends Message<T> {
+  /**
+   * Identifies this message as a PUBLISH packet. MQTTMessages created with
+   * the `binary` and `structured` Serializers will contain a "Content Type"
+   * property in the PUBLISH record.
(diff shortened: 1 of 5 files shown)


src/message/mqtt/index.ts
```diff
@@ -0,0 +1,114 @@
...
+ Copyright 2021 The CloudEvents Authors
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+import { Binding, Deserializer, CloudEvent, CloudEventV1, CONSTANTS, Message, ValidationError, Headers } from "../..";
+
+export {
+  MQTT,
+  MQTTMessage,
+};
```
grant, line 11: Do you want to add comments to exported symbols?

test/integration/mqtt_tests.ts
```diff
@@ -0,0 +1,309 @@
...
+};
+
+const ext1Name = "extension1";
+const ext1Value = "foobar";
+const ext2Name = "extension2";
+const ext2Value = "acme";
+
+// Binary data as base64
+const dataBinary = Uint32Array.from(JSON.stringify(data), (c) => c.codePointAt(0) as number);
+const data_base64 = asBase64(dataBinary);
```
grant, line 36: nit: it may be better to keep/group related constants in an object rather than having a bunch of individual consts.

grant (approved):
Fine with me :)

Might want to add some more comments for readability and better IDE devX.

PR #465, opened 2022-01-24
chore: update readme with current Node LTS versions

lholmquist:


lance (approved):
We should add 16.x to our test matrix for the GH action!

Issue #456, opened 2021-12-29
Can not build with --isolatedModules


neelance:
**Describe the Bug**
Can not build with [TypeScript compiler option](https://www.typescriptlang.org/tsconfig#isolatedModules) `isolatedModules`.

**Steps to Reproduce**
Add `cloudevents` as a dependency to a project that uses `isolatedModules` and run a build.

You get:
```
node_modules/cloudevents/dist/index.d.ts:7:60 - error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.
```

The problem is with `export declare const enum Version`. One way to solve this would be to drop the `const` modifier.

**Expected Behavior**
Build passes.

jketcham:
I'm running into this issue as well in a Typescript project being transpiled with babel.

For reference, it looks like there are two places where `const enum` is being used:
- https://github.com/cloudevents/sdk-javascript/blob/2ac731eb884965e91a19bb3529100a6aee7069dd/src/event/cloudevent.ts#L17
- https://github.com/cloudevents/sdk-javascript/blob/80d987c1f6046efb5e0c89b0472d653ccd35ee2c/src/transport/protocols.ts#L10

Could a change to remove the `const` declaration from those enums be accepted? 

Or I think setting `preserveConstEnums` to `true` may fix the problem too, info on that option here: https://www.typescriptlang.org/tsconfig#preserveConstEnums.

I tried both changes in a fork and all tests passed if anyone is interested in taking a look. I can open a PR if either change is reasonable.
- https://github.com/jketcham/sdk-javascript/tree/456-remove-const-enums
- https://github.com/jketcham/sdk-javascript/tree/456-set-preserve-const-enums

grant:
@lance Do you think we could simply remove `const` from our enums here?

Seeing a downstream issue.

lance:
> @lance Do you think we could simply remove `const` from our enums here?

That should be part of it. But I think the resolution should be more comprehensive. I found this article a good read on the problem, as I was unaware of exactly why this was an issue. https://ncjamieson.com/dont-export-const-enums/

We should add `"isolatedModules": "true"` to our tsconfig.json. I did this locally, and it actually exposes a number of problems besides just this.

```
❯ npm run build

> cloudevents@5.3.1 build /home/lanceball/src/github.com/cloudevents/sdk-javascript
> tsc --project tsconfig.json && tsc --project tsconfig.browser.json && webpack

[...]

PR #469, opened 2022-02-09
fix: use `isolatedModules: true` in tsconfig.json

lance:
This setting ensures the module can be used in projects where the workflow
includes type checking and transpilation as two separate steps.

See: https://ncjamieson.com/dont-export-const-enums/

Fixes: https://github.com/cloudevents/sdk-javascript/issues/456

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,9 +16,15 @@ import CONSTANTS from "./constants";
 
 export {
   // From event
-  CloudEvent, Version,
-  ValidationError, Mode, HTTP,
-  Kafka, MQTT, MQTTMessageFactory, emitterFor,
+  CloudEvent,
+  Version,
+  ValidationError,
+  Mode,
+  HTTP,
+  Kafka,
+  MQTT,
+  MQTTMessageFactory,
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -16,31 +16,24 @@ import CONSTANTS from "./constants";
 
 export {
   // From event
-  CloudEvent,
+  CloudEvent, Version,
```
grant, line 19: I would keep identitiers on separate lines. Here and elsewhere.

grant (approved):
Formatting changes should be separate. Otherwise lgtm

src/index.ts
lance: These formatting changes were auto applied by vscode. Perhaps we should have a lint rule?

lance:
@jketcham we're pretty loose about the timing of things. I'd like to land https://github.com/cloudevents/sdk-javascript/pull/471 before another release, but assuming that's relatively non-controversial, I'd imagine we could have a release out next week at the latest.

jketcham:
@lance sounds good 👍

Issue #423, opened 2021-07-28
AJV schema validation functions should be pre-compiled to avoid unsafe-eval CSP requirement


hrobertson:
**Describe the Bug**
Compiling the AJV schema validation functions at runtime uses new `new Function` which requires the CSP `script-src` to include `'unsafe-eval'`.

Instead, the schema validation functions should be pre-compiled.

**Steps to Reproduce**
1. Use the cloudevents/sdk-javascript library in a site with a CSP with `script-src` not including `'unsafe-eval'`
2. Use `new CloudEvent(...)`
3. Observe CSP error

**Expected Behavior**
The library should be able to be used successfully without including `'unsafe-eval'` in your CSP.

**Additional context**
See https://github.com/ajv-validator/ajv/blob/master/docs/security.md#content-security-policy

lance:
@hrobertson thanks for reporting this.

See also: https://github.com/ajv-validator/ajv/blob/master/docs/standalone.md

lance:
I have been working on this, but am awaiting some response from the ajv team https://github.com/ajv-validator/ajv/issues/1837

PR #471, opened 2022-02-10
feat: precompile cloudevent schema

lance:
This commit modifies the build pipleline so that the cloudevent schema is
precompiled for runtime validation. This eliminates the need to compile the
schema at runtime, improving both performance and security.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/423

~~Note: There are a number of changes related to semi-dead code from spec version 0.3 that did not get removed when we dropped support for that version.~~


src/event/spec.ts
```diff
@@ -3,28 +3,17 @@
...
-});
-
-const isValidAgainstSchemaV1 = ajv.compile(schemaV1);
 
 export function validateCloudEvent<T>(event: CloudEventV1<T>): boolean {
   if (event.specversion === Version.V1) {
-    if (!isValidAgainstSchemaV1(event)) {
-      throw new ValidationError("invalid payload", isValidAgainstSchemaV1.errors);
+    if (!validate(event)) {
+      throw new ValidationError("invalid payload", (validate as any).errors);
```
grant, line 16: `(validate as any)` seems a bit off. Why do we need to cast to `any`?

I'm seeing `validate` as a function:

```ts
function validate(data: any, { instancePath, parentData, parentDataProperty, rootData }?: {
    instancePath?: string | undefined;
    parentData: any;
    parentDataProperty: any;
    rootData?: any;
}): boolean
```

src/schema/cloudevent.json
```diff
@@ -0,0 +1,128 @@
...
+    "data_base64": {
+      "description": "Base64 encoded event payload. Must adhere to RFC4648.",
+      "$ref": "#/definitions/data_base64def",
+      "examples": [
+        "Zm9vYg=="
+      ]
+    }
+  },
+  "required": ["id", "source", "specversion", "type"],
+  "definitions": {
```
grant, line 84: I don't think the definitions need to have the suffix of `def`. Can you remove this suffix from the definitions?

src/schema/v1.js
```diff
@@ -0,0 +1 @@
..."}];return false;}}}}}var valid0 = _errs22 === errors;}else {var valid0 = true;}if(valid0){if(data.data !== undefined){let data8 = data.data;const _errs25 = errors;if((((typeof data8 != "object") && (typeof data8 !== "string")) && (!((typeof data8 == "number") && (isFinite(data8))))) && (typeof data8 !== "boolean")){validate20.errors = [{instancePath:instancePath+"/data",schemaPath:"#/definitions/datadef/type",keyword:"type",params:{type: schema31.type},message:"must be object,string,number,array,boolean,null"}];return false;}var valid0 = _errs25 === errors;}else {var valid0 = true;}if(valid0){if(data.data_base64 !== undefined){let data9 = data.data_base64;const _errs28 = errors;if((typeof data9 !== "string") && (data9 !== null)){validate20.errors = [{instancePath:instancePath+"/data_base64",schemaPath:"#/definitions/data_base64def/type",keyword:"type",params:{type: schema32.type},message:"must be string,null"}];return false;}var valid0 = _errs28 === errors;}else {var valid0 = true;}}}}}}}}}}}}else {validate20.errors = [{instancePath,schemaPath:"#/type",keyword:"type",params:{type: "object"},message:"must be object"}];return false;}}validate20.errors = vErrors;return errors === 0;}
```
grant, line 1: I don't think we should check in compiled JS like this in our source repo.

grant (changes requested):
Going to request a change at least for `src/schema/v1.js` to not be checked in, or to not be generated.

src/event/spec.ts
lance: The function that AJV generates has an `errors` object attached to it if there are errors found when validating with `validate(data)`. It's a non-optimal API. Here you can see it in the generated validation code. I welcome suggestions on how to deal with this another way.

```js
function validate20(data, { instancePath = "", parentData, parentDataProperty, rootData = data } = {}) {
  let vErrors = null;
  let errors = 0;
  if (errors === 0) {
    if (data && typeof data == "object" && !Array.isArray(data)) {
      let missing0;
      if (
        (data.id === undefined && (missing0 = "id")) ||
        (data.source === undefined && (missing0 = "source")) ||
        (data.specversion === undefined && (missing0 = "specversion")) ||
        (data.type === undefined && (missing0 = "type"))
      ) {
        validate20.errors = [
          {
            instancePath,
            schemaPath: "#/required",
            keyword: "required",
            params: { missingProperty: missing0 },
            message: `must have required property '${  missing0  }'`,
          },
        ];
        return false;
      }
```

src/schema/cloudevent.json
lance: This is taken directly from the CloudEvent specification here: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/cloudevents.json. I would prefer it if we can use this document unchanged.

src/event/spec.ts
```diff
@@ -3,28 +3,17 @@
...
-  return date.toString() !== "Invalid Date";
-});
-
-const isValidAgainstSchemaV1 = ajv.compile(schemaV1);
 
 export function validateCloudEvent<T>(event: CloudEventV1<T>): boolean {
   if (event.specversion === Version.V1) {
-    if (!isValidAgainstSchemaV1(event)) {
-      throw new ValidationError("invalid payload", isValidAgainstSchemaV1.errors);
+    if (!validate(event)) {
```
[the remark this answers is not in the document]
grant: Ah, alrighty then

src/schema/v1.js
lance: I understand your sentiment here and I considered not adding it to the PR. In most cases, you of course do not want to add generated source to revision control. However, there are some cases where it might make sense. For example, `package-lock.json` is a generated file, and it's considered best practice to include it in your repository. Many golang projects also include generated code in a repo. I think it's fair to say that not doing so is a guideline, and not a hard and fast rule.

In this case, I think it perhaps _does_ make sense to include it. The main reason is that it hardly ever changes. If we were using `Makefile`, the code would only ever be generated when `schema/cloudevent.json` is generated. Of course we are not using `Makefile` so it's always generated, but still... I like the fact that we will know that it has changed, since any change appears in git. It is also nice to be able to validate the change when necessary. If `schema/cloudevent.json` changes, then `schema/v1.js` MUST also change. We will know for sure that is the case, if and only if it's checked in.

I am willing to concede, but I definitely thought about it, and definitely made the conscious choice to include it for those reasons.

test/integration/constants_test.ts
```diff
@@ -36,64 +36,6 @@ describe("Constants exposed by top level exports", () => {
       `${CONSTANTS.MIME_CE_JSON}; charset=${CONSTANTS.CHARSET_DEFAULT}`,
     );
   });
-  describe("V0.3 binary headers constants", () => {
```
grant, line 39: This change seems separate from the rest of the PR. Can you have a separate PR for this change?

lance:
@grant requested changes have been made - ptal

grant (approved):
Thanks!

PR #483, opened 2022-03-18
docs: update readme to include http builtin transport

lance:
In addition to adding a simple example, I changed the `Emitter` singleton example to use this, as it's much simpler than the axios implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ app.post("/", (req, res) => {
 
 #### Emitting Events
 
-The easiest way to send events is to use the builtin HTTP emitter.
+The easiest way to send events is to use the built-in HTTP emitter.
 
 ```js
 const { httpTransport, emitterFor, CloudEvent } = require("cloudevents");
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,9 +46,26 @@ app.post("/", (req, res) => {
 
 #### Emitting Events
 
-You can send events over HTTP in either binary or structured format
-using the `HTTP` binding to create a `Message` which has properties
-for `headers` and `body`.
+The easiest way to send events is to use the builtin HTTP emitter.
```
grant, line 49: `builtin` -> `built-in`

README.md
```diff
@@ -14,7 +14,7 @@ The CloudEvents SDK for JavaScript.
 - Serialize and deserialize CloudEvents in different [event formats](https://github.com/cloudevents/spec/blob/v1.0/spec.md#event-format).
 - Send and recieve CloudEvents with via different [protocol bindings](https://github.com/cloudevents/spec/blob/v1.0/spec.md#protocol-binding).
 
-_Note:_ Supports CloudEvent versions 0.3, 1.0
```
grant, line 17: no 0.3? Doesn't the spec say we should support 0.3?

This change seems tangential to the rest of the change.

grant (approved):

README.md
lance: 0.3 support was removed last August in https://github.com/cloudevents/sdk-javascript/pull/425 but this line was missed in the removal. Support for 0.3 is considered optional, so it was removed in favor of simplicity. There may still be some unfortunate remnants around, like this, that need clean up.

Issue #491, opened 2022-06-02
Non Uint32Array binary data does not set the CloudEvent's data_base64 value


matthewchngshopback:
When binary data, e.. `Buffer` or `Uint8Array`, is not converted to `Uint32Array` before being set as a CloudEvent's data, the `data_base64` will not be set because `isBinary()` detection fails. This results in **structured mode** to go out of spec by not having the expected `data_base64` attribute but instead include an unexpected `data` object.

If we convert the `Buffer` to `Uint32Array` and set as CloudEvent's data, then **binary mode** message body will contain wrong byte sequence as it will be that of `Uint32Array` rather than the expected `Uint8Array`. This can be overcome by wrapping the `body` from `HTTP.binary()` with `Buffer.from(body)` again before feeding it `axios()`. Also, **`Uint32Array` binary data in binary mode** does not work with the provided `emitterFor()` as the `req.write()` only allows `Buffer` or `Uint8Array`.

**Steps to Reproduce**
```
const { CloudEvent, HTTP } = require("cloudevents")

const x = Buffer.from([0x01, 0x02, 0x03, 0x4])
const ce = new CloudEvent({ type: 'test', source:'testor', data: x })
const { body } = HTTP.structured(ce)

console.log(`ce: ${ce}`)
// ce: {"id":"084afd51-9dd2-4f0c-8f4f-517189d6ca13","time":"2022-06-02T07:34:30.982Z","type":"test","source":"testor","specversion":"1.0","data":{"type":"Buffer","data":[1,2,3,4]}}

console.log(`structured body: ${body}`)
// structured body: {"id":"084afd51-9dd2-4f0c-8f4f-517189d6ca13","time":"2022-06-02T07:34:30.982Z","type":"test","source":"testor","specversion":"1.0","data":{"type":"Buffer","data":[1,2,3,4]}}

const x32 = new Uint32Array(x)
const ceX32 = new CloudEvent({ type: 'test', source:'testor', data: x32 })
const { body: bodyX32 } = HTTP.structured(ceX32)

console.log(`ceX32: ${ceX32}`)
// ceX32: {"id":"ca9fb24f-adc1-4c8a-82e2-ff0f69b26cbd","time":"2022-06-02T07:38:27.833Z","type":"test","source":"testor","specversion":"1.0","data_base64":"AQIDBA==","data":{"0":1,"1":2,"2":3,"3":4}}

console.log(`structured bodyX32: ${bodyX32}`)
[...]

PR #494, opened 2022-06-14
fix: allow `TypedArray` for binary data

lance:
Previously we only considered `Uint32Array` binary data. This was an oversight. This fixes that issue.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/491

### DIFF
--- a/src/event/validation.ts
+++ b/src/event/validation.ts
@@ -36,7 +36,10 @@ export const isDefined = (v: unknown): boolean => v !== null && typeof v !== "un
 export const isBoolean = (v: unknown): boolean => typeof v === "boolean";
 export const isInteger = (v: unknown): boolean => Number.isInteger(v as number);
 export const isDate = (v: unknown): v is Date => v instanceof Date;
-export const isBinary = (v: unknown): v is Uint32Array => v instanceof Uint32Array;
+export const isBinary = (v: unknown): boolean =>
+  v instanceof Uint32Array ||
+  v instanceof Uint16Array ||
+  v instanceof Uint8Array;
 
 export const isStringOrThrow = (v: unknown, t: Error): boolean =>
   isString(v)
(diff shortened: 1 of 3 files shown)


test/integration/spec_1_tests.ts
```diff
@@ -182,5 +182,25 @@ describe("CloudEvents Spec v1.0", () => {
       const ce = cloudevent.cloneWith({ datacontenttype: "text/plain", data: dataBinary });
       expect(ce.data_base64).to.equal(expected);
     });
+
+    it("should be ok when type is 'Uint16Array' for 'Binary'", () => {
```
grant, line 186: I would just have 3 `expect` calls for each Uint array type in 1 test.

src/event/validation.ts
```diff
@@ -36,7 +36,10 @@ export const isDefined = (v: unknown): boolean => v !== null && typeof v !== "un
 export const isBoolean = (v: unknown): boolean => typeof v === "boolean";
 export const isInteger = (v: unknown): boolean => Number.isInteger(v as number);
 export const isDate = (v: unknown): v is Date => v instanceof Date;
-export const isBinary = (v: unknown): v is Uint32Array => v instanceof Uint32Array;
+export const isBinary = (v: unknown): boolean =>
+  v instanceof Uint32Array ||
+  v instanceof Uint16Array ||
+  v instanceof Uint8Array;
```
grant, line 42: I think there must be a better way to do this than to hardcode all typed arrays.

What about `ArrayBuffer.isView`?

src/event/validation.ts
```diff
@@ -82,7 +85,8 @@ export const asBuffer = (value: string | Buffer | Uint32Array): Buffer =>
         throw new TypeError("is not buffer or a valid binary");
       })();
 
-export const asBase64 = (value: string | Buffer | Uint32Array): string => asBuffer(value).toString("base64");
+export const asBase64 = 
+(value: string | Buffer | Uint32Array | Uint16Array | Uint8Array): string => asBuffer(value).toString("base64");
```
grant, line 89: Is this type exhaustive of all types? Maybe we can use a separate typed array `type` here.

https://stackoverflow.com/a/51681541

grant (changes requested):

lance:
Updated with your suggestions @grant and CI is passing. PTAL

grant (approved):
LGTM

Issue #489, opened 2022-05-13
bug: parsing big integers in JSON `data` loses precision


lance:
**Describe the Bug**
When parsing JSON data, if a JSON field value is a number, and that number is really big, JavaScript loses precision. For example, the Twitter API exposes the Tweet ID. This is a large number that exceeds the integer space of `Number`. You can see this by simply assigning a large number to a simple variable (it's not actually JSON that is the problem, it's JavaScript itself).

```
> let id = 1524831183200260097
undefined
> id
1524831183200260000
```

**Steps to Reproduce**
```
❯ npm install cloudevents
❯ cat > index.js
const { CloudEvent } = require('cloudevents')

let e = new CloudEvent({ source: 'example', type: 'example', data: 1524831183200260097 })
console.log(e.data)
^D
❯ node index.js
1524831183200260000
```

**Expected Behavior**
I expect the data to not lose precision.

**Additional context**
This can be resolved by using `json-bigint`

```
> let JSON = require('json-bigint')({useNativeBigInt: true})
> let j = JSON.parse('{ "key": 993143214321423154315154321 }')
> j.key
993143214321423154315154321
```

cardil:
I think your _Steps to Reproduce_ are a little misleading. You type there in verbatim a long number. JavaScript engine, parse a file, and use [IEEE 754 Standard](https://en.wikipedia.org/wiki/IEEE_754) to parse a `Number`. You could avoid that by using `BigInt` notation:

```js
let e = new CloudEvent({ source: 'example', type: 'example', data: 1524831183200260097n })
```

The problem exists, though. The problem lies in parsers implemented in this library. Here's simple jest for `HTTP`:

```js
const { HTTP } = require('cloudevents')

test('Parse big numbers from HTTP', () => {
  let ce = HTTP.toEvent({
    headers: { 'content-type': 'application/cloudevents+json' },
    body: `{"id": 1524831183200260097}`
  })
  expect(ce.id).toBe(1524831183200260097n);
})
```

This test fails:

```
  ● Parse big numbers from HTTP

    expect(received).toBe(expected) // Object.is equality

    Expected: 1524831183200260097n
    Received: 1524831183200260000

       6 |     body: `{"id": 1524831183200260097}`
       7 |   })
    >  8 |   expect(ce.id).toBe(1524831183200260097n)
         |                 ^
       9 | })
      10 |
      11 |

      at Object.toBe (bigint.test.js:8:17)
```

[...]

PR #495, opened 2022-06-14
fix: handle big integers in incoming events

lance:
An event may have data that contains a BigInt. The builtin `JSON` parser
for JavaScript does not handle the `BigInt` types. The introduced
`json-bigint` dependency (34k) does.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/489

### DIFF
--- a/src/message/http/index.ts
+++ b/src/message/http/index.ts
@@ -4,6 +4,7 @@
 */
 
 import { types } from "util";
+import JSONbig from "json-bigint";
 
 import { CloudEvent, CloudEventV1, CONSTANTS, Mode, Version } from "../..";
 import { Message, Headers, Binding } from "..";
@@ -18,6 +19,8 @@ import {
 } from "./headers";
 import { isStringOrObjectOrThrow, ValidationError } from "../../event/validation";
 import { JSONParser, MappedParser, Parser, parserByContentType } from "../../parsers";
+
+const JSON = JSONbig(({ useNativeBigInt: true }));
 
 /**
  * Serialize a CloudEvent for HTTP transport in binary mode
--- a/src/message/http/index.ts
+++ b/src/message/http/index.ts
@@ -4,7 +4,6 @@
 */
 
 import { types } from "util";
-import JSONbig from "json-bigint";
 
 import { CloudEvent, CloudEventV1, CONSTANTS, Mode, Version } from "../..";
 import { Message, Headers, Binding } from "..";
@@ -19,8 +18,6 @@ import {
 } from "./headers";
 import { isStringOrObjectOrThrow, ValidationError } from "../../event/validation";
 import { JSONParser, MappedParser, Parser, parserByContentType } from "../../parsers";
-
-const JSON = JSONbig(({ useNativeBigInt: true }));
 
 /**
  * Serialize a CloudEvent for HTTP transport in binary mode
--- a/src/parsers.ts
+++ b/src/parsers.ts
@@ -3,9 +3,11 @@
  SPDX-License-Identifier: Apache-2.0
 */
 
+import JSONbig from "json-bigint";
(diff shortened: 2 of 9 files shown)


grant (approved):

src/parsers.ts
```diff
@@ -3,6 +3,7 @@
  SPDX-License-Identifier: Apache-2.0
 */
 
+import JSON from "json-bigint";
```
cardil, line 6:
```suggestion
const JSON = require('json-bigint')({ useNativeBigInt: true });
```

I think we should use the native [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) JS type, by using the above import option.

Without it, the [bignumber.js](https://github.com/MikeMcl/bignumber.js) type is being used. It prevents users from simple comparisons, like native `BitInt` allows (note the lack of String casting):


```js
expect(ce.data).to.equal(1524831183200260097n);
```

src/message/http/index.ts
```diff
@@ -4,6 +4,7 @@
 */
 
 import { types } from "util";
+import JSON from "json-bigint";
```
cardil, line 7: Have you measured the impact of using this library versus using a native V8 implementation?

If the difference is significant (it might be), we should consider gating this behavior with a configuration option (it may be turned on by default IMHO).

cardil (changes requested):
Thanks a lot @lance for fixing this. It looks good, but I found some doubts about this proposal.

lance:
Closing and reopening to trigger re-run of CI with updated package.json (to hopefully fix errors)

src/parsers.ts
lance: Thanks for pointing this out. Much nicer.

src/message/http/index.ts
lance: This is a good point, so I decided to try a simple, probably naive test. But no matter how naive, you can't ignore a 7x slowdown with `bigint` support. 

```
❯ time node index.js
Unsing native JSON
....................................................................................................
node index.js  199.67s user 1.29s system 98% cpu 3:23.76 total

❯ time node index.js bigint
Using bigint
....................................................................................................
node index.js bigint  1308.75s user 4.49s system 99% cpu 21:55.06 total

❯ cat index.js
   1   │ const fs = require('fs');
   2   │
   3   │ const J = require('json-bigint')({ useNativeBitInt: true });
   4   │ let Parser;
   5   │
   6   │ if (process.argv[2] === 'bigint') {
   7   │   Parser = J;
   8   │   console.log('Using bigint');
   9   │ } else {
  10   │   Parser = JSON;
  11   │   console.log('Unsing native JSON');
  12   │ }
  13   │
  14   │ const json = fs.readFileSync('./generated.json');
  15   │
  16   │ let remain = 10_000_000;
  17   │ while (remain-- > 0) {
  18   │   Parser.parse(json);
  19   │   if (remain%100000 === 0) {
  20   │     process.stdout.write('.');
  21   │   }
  22   │ }
  23   │
  24   │ console.log();
```

There is no obvious place to set global parser variables. I need to think on this a bit.

cardil: @lance Have you thought of where to place this feature flag?

lance:
@cardil I have added an environment variable to enable the `json-bigint` module. I'm not a huge fan of this approach because the check is made for each call to the parser. The library does not have an obvious place to set flags or other static values. An alternative approach would be to add this as an option to the `HTTP.toEvent()` method, but that changes the method signature resulting in a lot of small changes all over the codebase.

@lholmquist ptal

lholmquist (approved):
/lgtm. 

I do like your suggestion of adding this to the method signature, which as you said is a bigger change,  but maybe we can do that later then

Issue #515, opened 2022-10-30
Avoid jest#13535 and planttheidea/fast-equals#91


PSanetra:
**Describe the Bug**
`CloudEvent` class instances are at the moment not comparable using some popular libraries. Instances might be considered equal although they are not equal.

See the following issues: 
* https://github.com/facebook/jest/issues/13535
* https://github.com/planttheidea/fast-equals/issues/91

**Steps to Reproduce**
1. Create two CloudEvent instances with different data properties
2. Pass the instances to `expect(...).toStrictEqual(...)` of jest and to `circularDeepEqual(...)` of the fast-equals library
3. Both calls will consider those CloudEvent instances to be equal although they are not equal

**Expected Behavior**
Both instances should not be considered equal using those libraries.

**How to fix**
I think the best way to avoid this issue is to avoid using private javascript fields and use private Typescript fields instead.

PR #516, opened 2022-10-31
fix: Make CloudEvent data field immutable and enumerable

PSanetra:
## Proposed Changes
* CloudEvent class:
  * Make data field immutable like the other fields in the class
  * Make data field enumerable via `Object.keys()` and `Object.entries()`
  * toJSON() does not return a `data` field anymore if `data_base64` is set to comply with the JSON format spec 3.1.1 `Out of this follows that the presence of the data and data_base64 members is mutually exclusive in a JSON serialized CloudEvent.`

Fixes https://github.com/cloudevents/sdk-javascript/issues/515

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ typings/
 
 # Package lock
 package-lock.json
+
+# Jetbrains IDE directories
+.idea
--- a/examples/express-ex/index.js
+++ b/examples/express-ex/index.js
@@ -28,11 +28,11 @@ app.post("/", (req, res) => {
     const responseEventMessage = new CloudEvent({
       source: '/',
       type: 'event:response',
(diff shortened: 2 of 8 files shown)


lance (commented):
@PSanetra thanks for this contribution! The `atob` function does not exist in Node.js LTS prior to version 16. Could you please add `atob` to `dependencies` and `@types/atob` to the `dev-dependencies` section, and add an import statement in `validation.ts`? This should get CI passing.

PSanetra:
@lance I have changed the code to support node.js runtime and browser runtime without a additional dependency.

lance (approved):
Great addition. Thank you.

lholmquist (approved):

PR #526, opened 2023-02-10
fix: add `process` plugin for browser compatibility

### LINKED ISSUE #525: Does not work in browsers as of 5.3.1
**Describe the Bug**
Based on other issues in this repo and some of the examples, it looks like this package is expected to work in the browser. Using any version after 5.3.0, there is an error in the browser console with the text "process is not defined".

**Steps to Reproduce**
1. Create website project (e.g., `npm create vite@latest`)
2. Install any version of cloudevents > 5.3.0
3. Add `import { CloudEvent } from "cloudevents"` to a javascript file (and use it to ensure tree shaking doesn't remove the code).
4. Run the application
5. Note in the browser console, the error "process is not defined"

The issue appears to be caused by the `util` dependency introduced in 5.3.1.

**Expected Behavior**
cloudevents NPM package should work inside the browser.

lholmquist:
Adds `process` plugin for webpack 5 and greater.

fixes #525

see https://github.com/browserify/node-util/issues/57#issuecomment-764436352 for the fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "ajv": "^8.11.0",
         "ajv-formats": "^2.1.1",
+        "process": "^0.11.10",
         "util": "^0.12.4",
         "uuid": "^8.3.2"
       },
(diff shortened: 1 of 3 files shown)


lance (approved):
Thanks @lholmquist!

Issue #527, opened 2023-02-10
Add the engines property to the package.json


lholmquist:
We currently don't have an engines property in the package.json that defines the node versions this module supports.

Current Active versions of Node or 18 and 16.   Node 14 is in maintenance and goes EOL April 30th. Node 20 comes out in April.

lholmquist:
added via #529 and https://github.com/cloudevents/sdk-javascript/commit/bc3aaca2ef250e4acd72b909488b326233237c83

PR #529, opened 2023-02-11
chore: added the engines property to the package.json

viraj-s15:
Signed-off-by: viraj-s15 <veer15102003@gmail.com>

## Proposed Changes
Added the engines property to the package.json
## Description
Fixes: https://github.com/cloudevents/sdk-javascript/issues/527

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,6 +157,6 @@
   },
   "types": "./dist/index.d.ts",
   "engines": {
-    "node": ">=14.0.0 <20.0.0"
+    "node": ">=12 <20.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -155,5 +155,8 @@
   "publishConfig": {
     "access": "public"
   },
-  "types": "./dist/index.d.ts"
+  "types": "./dist/index.d.ts",
+  "engines": {
+    "node": ">=14.0.0 <20.0.0"
+  }
 }


package.json
```diff
@@ -155,5 +155,8 @@
   "publishConfig": {
     "access": "public"
   },
-  "types": "./dist/index.d.ts"
+  "types": "./dist/index.d.ts",
+  "engines": {
+    "node": ">=14.0.0 <20.0.0"
```
lance, line 160:
```suggestion
    "node": ">=12 <20.0.0"
```
We actually support v12.x and up

viraj-s15: Ahh I apologise. Reading the previous comment got me a little confused.

package.json
```diff
@@ -155,5 +155,8 @@
   "publishConfig": {
     "access": "public"
   },
-  "types": "./dist/index.d.ts"
+  "types": "./dist/index.d.ts",
+  "engines": {
+    "node": ">=12 <20.0.0"
```
lholmquist, line 160: Node 12 is EOL'd, but for us to drop it, that would require a major version bump,  which we can do in a separate PR,  so this is good.  

Not really a blocker,  but my preference would be to only have ">=12" and not the "<20.0.0"

lholmquist (approved):
Just need to sign the commit,  but LGTM

lholmquist:
closed via https://github.com/cloudevents/sdk-javascript/commit/bc3aaca2ef250e4acd72b909488b326233237c83

Issue #540, opened 2023-04-28
Getting error when installing on node 20.0.0


sskular-wonderlab:
**Describe the Bug**
When installing the package for my latest node version, 20.0.0., which became generally available on the 2023-04-18, I am getting an error message:

npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'cloudevents@6.0.4',
npm WARN EBADENGINE   required: { node: '>=12 <20.0.0' },
npm WARN EBADENGINE   current: { node: 'v20.0.0', npm: '9.6.5' }
npm WARN EBADENGINE }

**Steps to Reproduce**
1. npm i cloudevents

**Expected Behavior**
Package installation successfully finalises

**Additional context**
20.0.0 version became GA only recently, see https://github.com/cloudevents/sdk-javascript/blob/main/package.json#enroll-beta

So https://github.com/cloudevents/sdk-javascript/blob/main/package.json#enroll-beta should be updated to include 20.0.0 version once it's been verified cloudevents works with the new node version

PR #541, opened 2023-04-28
540 fix node version

sskular-wonderlab:
## Proposed Changes
Update Node version to 20.0.0

## Description
Fixes: https://github.com/cloudevents/sdk-javascript/issues/540

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,6 +158,6 @@
   },
   "types": "./dist/index.d.ts",
   "engines": {
-    "node": ">=12 <20.0.0"
+    "node": ">=12 <=20.0.0"
   }
 }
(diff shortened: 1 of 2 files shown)


lance (approved):
/lgtm

package.json
```diff
@@ -158,6 +158,6 @@
   },
   "types": "./dist/index.d.ts",
   "engines": {
-    "node": ">=12 <20.0.0"
+    "node": ">=12 <=20.0.0"
```
w3nl, line 161: Isn't it better to change it to `<=20`
Otherwise you say it's only working on 20.0.0, but not in 20.1.0 and 20.2.0, etc.

lance: Yes - that was addressed in #545 

https://github.com/cloudevents/sdk-javascript/blob/11442d32d307a0e8416ed573ce34fc825d3b63c4/package.json#L163

lance: Oh I see now... gotcha

PR #545, opened 2023-05-03
feat!: remove node 12 and node 14

lholmquist:
Node 12 has been EOL since the end of April 2022 and Node 14 just became EOL at the end of April 2023

This is a major version bump


lholmquist:
@lance ^^

PR #552, opened 2023-05-26
chore: Update compatible node version

w3nl:
## Proposed Changes

Allow Node 20.x

## Description

Now you allow only Node 20.0.0, let's make it compatible with all Node 20.x.x versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,6 +160,6 @@
   },
   "types": "./dist/index.d.ts",
   "engines": {
-    "node": ">=16 <=20.0.0"
+    "node": ">=16 <=20"
   }
 }


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cloudevents",
-  "version": "7.0.0",
+  "version": "7.0.1",
```
lance, line 3: This should stay at 7.0.0
```suggestion
  "version": "7.0.0",
```

pieter-pon: I've changed it back

lance: @pieter-pon thanks - you also should revert package-lock.json.

pieter-pon: @lance Also revert the package-lock, should I update the lock file that only the change for the engine is in the lock file, just like the normal package.json?

w3nl: @lance Also revert the package-lock, should I update the lock file that only the change for the engine is in the lock file, just like the normal package.json?

lholmquist (approved):

Issue #323, opened 2020-08-21
TypeError: Cannot read property 'V03' of undefined


moltar:
**Describe the Bug**

`Version` appears to be `undefined`.

**Steps to Reproduce**

```ts
import { CloudEvent, Version } from 'cloudevents'

new CloudEvent({
      id: event.id.toString(),
      source: event.source,
      type: event.type,
      subject: event.subject,
      time: event.time,
      data: event.data,
      specversion: Version.V03,
})
```

**Expected Behavior**

Version should be defined.

**Additional context**

Using package from NPM version `3.1.0`.

moltar:
Never mind, seems to be an issue with `@swc-node/jest`.

moltar:
See https://github.com/swc-project/swc/issues/940

Maybe this can refactored from enum to const as const approach?

```ts
const Versions  = {
  V03: 3,
} as const
```

lance:
Hi @moltar - I am unable to reproduce this. I tried downloading the `swc` CLI, and ran it on `src/event/cloudevent.ts` where the `Version` enum is declared, and did not get any errors. It produced this transpiled JS:

```js
export var Version;
(function(Version1) {
    Version1["V1"] = "1.0";
    Version1["V03"] = "0.3";
})(Version || (Version = {
}));
```
Can you provide some steps to reproduce this?

moltar:
Here is a repro: https://github.com/moltar/cloudevent-exp

lance:
Thanks @moltar. This really seems to be an issue with swc. By modifying your example to use ts-jest instead of swc the example runs without error.

```log
 npm t

> cloudevent-exp@1.0.0 test /home/lanceball/src/github.com/tmp/cloudevent-exp
> jest

 PASS  ./cloudevents.test.ts
  Version
    ✓ should work (2 ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        2.39 s
Ran all test suites.
```

My modifications.
- package.json
```diff
diff --git a/package.json b/package.json
index 2342abf..fa3a0b7 100644
--- a/package.json
+++ b/package.json
@@ -11,10 +11,12 @@
   "license": "ISC",
   "dependencies": {
     "@swc-node/jest": "0.1.27",
-    "cloudevent": "0.6.0",
     "cloudevents": "3.1.0",
     "jest": "26.4.2",
-    "ts-jest": "26.3.0",
[...]

PR #561, opened 2023-07-19
BREAKING CHANGE: use string instead of enum for `Version`

lance:
TypeScript does not consider enum values equivalent, even if the string representation is the same. So, when a module imports `cloudevents` and also has a dependency on another module which also has a dependency on `cloudevents` this can cause conflicts where the `CloudEvent.version` attribute is not considered equal when, in fact, it is.

Changing the `enum` to a string is pretty straightforward, but should be considered a breaking change since TypeScript dependents will potentially fail the build with a change like this.

See: https://github.com/knative/func/pull/1873#issuecomment-1640866356
Fixes: https://github.com/cloudevents/sdk-javascript/issues/323

### DIFF
--- a/src/event/cloudevent.ts
+++ b/src/event/cloudevent.ts
@@ -12,12 +12,10 @@ import { validateCloudEvent } from "./spec";
 import { ValidationError, isBinary, asBase64, isValidType, base64AsBinary } from "./validation";
 
 /**
- * An enum representing the CloudEvent specification version
+ * Constants representing the CloudEvent specification version
  */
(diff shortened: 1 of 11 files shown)


lholmquist (approved):
LGTM

lance:
> maybe that "empty commit release as version" workflow, that we've done before?

That didn't seem to work either... Hmm

lholmquist:
maybe the prefix of the commit should just be `fix!: .......`? I know that has worked in other repos

PR #573, opened 2023-12-04
chore: Update compatible node version

tnunamak:
## Proposed Changes
This indicates compatibility with Node.js 21.


lholmquist (changes requested):
could you also add node 21 to the CI matrix https://github.com/cloudevents/sdk-javascript/blob/main/.github/workflows/nodejs-ci-action.yml#L18

lholmquist (approved):
LGTM.  just need to sign that last commit to pass that DCO check

PR #593, opened 2024-07-22
fix: creating an event does not error when the event attribute name is too long

Cali0707:
From the spec, the requirement for the length of attribute names is only a `SHOULD`, not a `MUST`. Currently, if the length is over 20 then the sdk throws an error, which I believe is incorrect because the length requirement is not a `MUST`.

## Proposed Changes
- When an event attribute has a name lonoger than 2o characters, the SDK does not throw an error anymore


lholmquist (approved):
LGTM.

Cali0707:
Thanks @lholmquist !!

Issue #487, opened 2022-04-14
Return type definition of Kafka.structured seems incorrect


evanshortiss:
**Describe the Bug**

The types for return values from `Kafka.structured(cloudEventInstance)` seem to be incorrect. The returned object has the following structure if viewed using logs or a debugger:

```js
{
  key: 'fubar',
  value: '{"id":"ef233aa0-00a0-4b5a-9886-f8ff7566ef50","time":"2022-04-14T20:26:32.291Z","type":"foo","source":"bar","specversion":"1.0","partitionkey":"fubar","data":{"hello":"world}}',
  headers: { 'content-type': 'application/cloudevents+json; charset=utf-8' },
  body: '{"id":"ef233aa0-00a0-4b5a-9886-f8ff7566ef50","time":"2022-04-14T20:26:32.291Z","type":"foo","source":"bar","specversion":"1.0","partitionkey":"fubar","data":{"hello":"world}}',
  timestamp: '1649967992291'
}
```

However, the types come from `CE.Message<string>`. This results in the following intellisense structure being suggested:

```js
{
  body: unknown,
  headers: CE.Headers
}
```

It seems as though the headers key is correct, but the body is incorrect, and other keys are missing.

**Steps to Reproduce**
1. Create a CloudEvent, e.g `new ce.CloudEvent({ type: 'foo', source: 'bar', data: { hello: 'world' }, partitionkey: 'foobar' }`
2. Pass it to, e.g `Kafka.structured(theEvent)`
3. Attempt to access the timestamp, value, or key. These are not known properties on the return value.
4. Try to use the body and you find it is incorrectly typed as `unknown`

**Expected Behavior**

I would have expected the types to reflect the Object in the *Describe the bug* section, e.g:

```ts
{
  key: string|undefined
  value: string,
  headers: CE.Headers,
  body: string,
  timestamp: string
}
``` 

Is this an invalid assumption to make? Passing the result of `Kafka.structured(cloudEventInstance)` object to a Kafka client such as KafkaJS requires casting the object to access the necessary properties to satisfy the TS compiler. It seems like this casting should be unnecessary.

**Additional context**

I'm using version 6.0.1 of the cloudevents module. The image below might help provide some context.

<img width="773" alt="Screenshot 2022-04-14 at 1 45 23 PM" src="https://user-images.githubusercontent.com/1303687/163472386-a14d8d36-b11c-44e3-ae38-f1b81360e119.png">

[...]

PR #604, opened 2025-04-01
fix: add generics to `Binding` type

adtennant:
## Proposed Changes

* Add generics to the `Binding` type so that the typings of the values returned by `binary` and `structured` can be more exact. 
* Fix the typing on `KafkaMessage` so that `T` is authoritative of the type of `value`.

Fixes: https://github.com/cloudevents/sdk-javascript/issues/487

## Description

The typing changes are not perfect, but they are closer to reality and backward compatible. They should remove a decent chunk of `as` casting in client code.


lholmquist (approved):

lholmquist:
@adtennant Thanks for the contribution

PR #646, opened 2026-05-14
chore: version bump typescript, uuid, ... , fix failing tests and typing issues

### LINKED ISSUE #647: GHSA-w5hq-g745-h8pq
Dependency uuid has vulnerablity https://github.com/advisories/GHSA-w5hq-g745-h8pq

uuid: Missing buffer bounds check in v3/v5/v6 when buf is provided

Hi, i have a Snyk scan on my project that report a vulnerability on uuid package. Scored 6.3 (Medium level) by Snyk and 7.1 by CVSS (High level). https://security.snyk.io/vuln/SNYK-JS-UUID-16133035

kings1ay3r:
## Proposed Changes
- Version bump typescript, uuid
- Fix TypeScript type errors in validation and parsers
- Correct header serialization for objects
- Resolve date and payload validation in test suites

## Description
1. Proper Header Serialization (src/message/http/headers.ts)
 Previously, when an extension property was an object, it was improperly cast to a string via value as string. In JavaScript, this results in the exact string "[object Object]" being sent over the network, completely destroying the payload data. The fix properly uses JSON.stringify() when the header value is an object, preserving the data structure.

2. Type-Safe JSON Parsing (src/parsers.ts)
 The JSONParser.parse() method was incorrectly typed to return a string, even though JSON.parse() returns an Object/Array. It was corrected to return unknown (which is standard for parsed JSON in TypeScript), preventing downstream typing assumptions.

3. Test Predictability (test/conformance/steps.ts & test/integration/emitter_factory_test.ts)
 The conformance tests were blindly parsing missing dates, causing Invalid Date crashes if a time was missing. The fix safely checks if time exists before comparing date strings.
In emitter_factory_test.ts, the assertions now handle both pre-parsed objects and stringified JSON properties flexibly rather than blindly asserting deep equality on mismatched types.

4. Error Message Formatting (src/event/validation.ts)
 The validation Error constructor was heavily relying on a dirty hack (// @ts-ignore) to coerce arrays of error objects into a single string. This was refactored to check Array.isArray(errors) and use a type-safe reduce loop to convert ErrorObject validation failures into a legible string.

6. Dependency Bumps (package.json)
 DevDependencies like uuid, mocha, @cucumber/cucumber, axios, and typescript were suffering from legacy versioning that caused some of those implicit compilation errors you might have noticed. They were brought up to modern compatibility. 
 UUID package had a moderate vulnerability https://github.com/advisories/GHSA-w5hq-g745-h8pq
 ```
 uuid  
 Severity: moderate
 uuid: Missing buffer bounds check in v3/v5/v6 when buf is provided https://github.com/advisories/GHSA-w5hq-g745-h8pq
```

Fixes https://github.com/cloudevents/sdk-javascript/issues/647

### CODE REVIEW

**Comment by kings1ay3r** [comment]
@lance @lholmquist can you please take a look?

**Comment by kings1ay3r** [comment]
@lance @lholmquist can you please take a look?

File: src/parsers.ts [hunk 1]
```diff
@@ -89,9 +89,9 @@ export interface MappedParser {
 
 export class DateParser extends Parser {
   parse(payload: string): string {
-    let date = new Date(Date.parse(payload));
+    const date = new Date(Date.parse(payload));
```
blackakula, line 92: Creating `Date` object is not needed when the date is invalid.
Better simply check `isNaN()` right after `Date.parse()`.
[Source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#return_value):
> If dateString fails to be parsed as a valid date, [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) is returned.

Suggested version of this class:
```typescript
export class DateParser extends Parser {
  parse(payload: string): string {
    const timestamp = Date.parse(payload);
    return (
      isNaN(timestamp)
        ? new Date()
        : new Date(timestamp)
    ).toISOString();
  }
}
```

src/message/http/headers.ts
```diff
@@ -41,7 +41,12 @@ export function headersFor<T>(event: CloudEventV1<T>): Headers {
       if (map) {
         headers[map.name] = map.parser.parse(value as string) as string;
       } else if (property !== CONSTANTS.DATA_ATTRIBUTE && property !== `${CONSTANTS.DATA_ATTRIBUTE}_base64`) {
-        headers[`${CONSTANTS.EXTENSIONS_PREFIX}${property}`] = value as string;
+        const headerName = `${CONSTANTS.EXTENSIONS_PREFIX}${property}`;
+        if (typeof value === "object") {
```
blackakula, line 45: This might be a breaking change for projects that are using the wrong behavior (with "[object Object]").

Also note that `typeof value === "object"` is not a reliable check in JavaScript. For example, this check will destroy the error data:
- 🔴 `JSON.stringify(new Error('useful message'))`
- 🟢 `String(new Error('useful message'))`

blackakula:
The only usage of `uuid` in [src/event/cloudevent.ts](https://github.com/cloudevents/sdk-javascript/blob/main/src/event/cloudevent.ts#L58) should be compatible with the latest version.
Therefore, vulnerability [GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq) can be fixed in end-user projects by overriding its version in package manager. Example for `pnpm-workspace.yaml`:
```yaml
overrides:
  'uuid@8.3.2': '14.0.0'
```

PR #650, opened 2026-08-17
refactor: drop the @ts-ignore from ValidationError

kahirokunn:
## Proposed Changes

Remove the `@ts-ignore` from `ValidationError`.

## Description

Type the local error array as `Array<string | ErrorObject>` so TypeScript can check the `reduce()` call without changing runtime behavior. Focused tests cover string, Ajv, and missing errors.


kahirokunn:

kahirokunn:
keep

PR #651, opened 2026-08-17
docs: return the transport promise from custom emitter examples

kahirokunn:
## Proposed Changes

Return the Axios promise in the custom emitter examples.

## Description

Returning the promise lets callers wait for Axios with `await emit(event)`.


kahirokunn:

kahirokunn:
keep

PR #654, opened 2026-08-17
feat!: use Fetch API for HTTP transport

### LINKED ISSUE #534: Can't resolve 'http' error when using on React app
**Describe the Bug**
I'm trying to use CE javascript SDK in a React app. I have the following err:
```
ERROR in ./node_modules/cloudevents/dist/transport/http/index.js 16:31-46

Module not found: Error: Can't resolve 'http' in '/home/ksuszyns/git/openshift-knative-showcase/frontend/node_modules/cloudevents/dist/transport/http'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "http": require.resolve("stream-http") }'
	- install 'stream-http'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "http": false }
```

That error is thrown just after trying to create the CE:

```ts
const ce = new CloudEvent({})
```

I just wanted to parse incoming Server Sent Events data that is in structured-mode CE.

**Steps to Reproduce**
1. Create React app with: `npx create-react-app frontend --template typescript`
2. Install CE: `npm i --save cloudevents`
3. Import CE and attempt to create basic object:
   ```ts
   import { CloudEvent } from 'cloudevents'

   const ce = new CloudEvent({})
   ```
You can also see this error on this commit: https://github.com/cardil/openshift-knative-showcase/commit/cc8ffe2ab26a5c6fec783f5e203f7e6953f3f7c7

**Expected Behavior**

I just want to create basic CE objects, in React app. And parse them from the JSON text, in the near future. No errors should be thrown.

### LINKED ISSUE #564: Add support for request timeout in HTTP transport
I'd be happy to propose a PR, but I'd recommend rewriting the transport with [fetch](https://nodejs.org/api/globals.html#fetch) instead of extending the current approach:

https://github.com/cloudevents/sdk-javascript/blob/db60220762745f95514d609a8450aa6230ca05fc/src/transport/http/index.ts#L49

... As I'd have to use `setTimeout()`, which would make the code more complicated.

The drawback of `fetch()` is that it'd require Node.js 18+. Which shouldn't be too bad as Node.js 16 is reaching EOL in a few weeks.

kahirokunn:
Fixes: https://github.com/cloudevents/sdk-javascript/issues/564
Fixes: https://github.com/cloudevents/sdk-javascript/issues/534

## Proposed Changes

Use the Fetch API for the built-in HTTP transport, add opt-in retries with `withRetry()`, and add configurable emitter timeouts with `withTimeout()`.

## Description

`httpTransport()` now resolves successful sends with the native Fetch `Response` and the handled body. Non-2xx responses, aborts, and network failures reject with `HTTPTransportError`. The transport accepts Fetch headers and options, response handlers, and transport-wide or per-send abort signals. Redirects fail by default unless enabled with `fetchOptions.redirect`.

`withRetry()` retries common temporary HTTP failures with capped randomized exponential backoff and honors `Retry-After` for 429 and 503 responses. Every attempt sends the same CloudEvent, so receivers should handle duplicates.

`withTimeout()` gives each emitter invocation a configurable deadline and combines it with a caller-provided abort signal. When composed inside `withRetry()`, each attempt receives a fresh timeout; when composed outside it, one timeout covers every attempt and retry backoff. Timed-out HTTP sends reject with an `HTTPTransportError` whose `kind` is `aborted`, and the default retry policy does not retry them.

See `API_TRANSITION_GUIDE.md` for the breaking changes and migration examples.

### DIFF
--- a/API_TRANSITION_GUIDE.md
+++ b/API_TRANSITION_GUIDE.md
@@ -100,3 +100,166 @@ new CloudEvent({ type, source, data }).emit();
 
 // You can also have several listener to send the event to several endpoint
 ```
+
+### Upgrading From 10.x to 11.0
+
+In the 11.0.0 release, the built-in HTTP transport was rewritten on top of the Fetch API.
+
+#### HTTP Transport
+
+`httpTransport()` used to resolve with the response for every status code, so a receiver
+that rejected an event looked just like one that accepted it.
+
+```js
+import { CloudEvent, emitterFor, httpTransport } from "cloudevents";
+
+const emit = emitterFor(httpTransport("https://my.receiver.com/endpoint"));
+
+// resolved with { body, headers } whether the receiver returned 202 or 503
+const response = await emit(new CloudEvent({ type, source, data }));
+```
+
+A 2xx response now resolves with the native Fetch `Response` and its body as text, and
+anything else rejects with an `HTTPTransportError`.
+
+```js
+import {
+  CloudEvent,
+  emitterFor,
+  HTTPTransportError,
+  httpTransport,
+} from "cloudevents";
+
+const emit = emitterFor(httpTransport("https://my.receiver.com/endpoint"));
+
+try {
+  const { response, body } = await emit(new CloudEvent({ type, source, data }));
+  console.log(response.status, body);
+} catch (error) {
+  if (error instanceof HTTPTransportError) {
+    console.error(error.kind, error.response?.status);
+  }
+  throw error;
+}
+```
+
+The transport takes a second argument now, for headers, a signal, how the response body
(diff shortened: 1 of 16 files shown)


kahirokunn:

kahirokunn:
@lholmquist Hi ✋ would you mind taking a look when you have a chance? Thank you 🙏

kahirokunn:
keep

kahirokunn:
keep

